PR #11, opened 2017-04-26
Jcs/database tests

jSylvestre:


Test/TestsDatabase/OrderDatabaseTests.cs
```diff
@@ -0,0 +1,173 @@
...
+using Test.Helpers;
+using Xunit;
+
+namespace Test.TestsDatabase
+{
+    [Trait("Category", "DatabaseTests")]
+    public class OrderDatabaseTests
+    {
+        [Fact]
+        public void test1()
```
srkirkland: @jSylvestre is there such a thing as beforeTest/afterTest w/ xUnit?  Seems like opening/closing the connection would be a good thing to do there and would save a lot of boilerplate

Test/TestsDatabase/OrderDatabaseTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            try
+            {
+                var options = new DbContextOptionsBuilder<ApplicationDbContext>()
+                    .UseSqlite(connection)
+                    .Options;
+
+                using (var context = new ApplicationDbContext(options))
+                {
+                    context.Database.EnsureCreated();
+                }
```
srkirkland: @jSylvestre what is the point of using multiple contexts?

jSylvestre: @srkirkland It was in the test example:
https://docs.microsoft.com/en-us/ef/core/miscellaneous/testing/sqlite

Test/TestsDatabase/OrderDatabaseTests.cs
jSylvestre: @srkirkland Yes, but:
Note 2: The xUnit.net team feels that per-test setup and teardown creates difficult-to-follow and debug testing code, often causing unnecessary code to run before every single test is run. For more information, see http://jamesnewkirk.typepad.com/posts/2007/09/why-you-should-.html.

srkirkland: @jSylvestre ok, so maybe just a helper method that creates the db and opens a context would work better

Test/TestsDatabase/OrderDatabaseTests.cs
srkirkland: @jSylvestre Try it without and see if it makes any difference.  Seems like extra work to me

Test/TestsDatabase/OrderDatabaseTests.cs
srkirkland: @jSylvestre something like a class that implements iDisposible and opens up a new connection and wraps everything in a try/catch/finally.

PR #15, opened 2017-05-01
Context helper

jSylvestre:
@srkirkland a helper like this?

Should the connection be outside the help so it can be closed in each test?

### DIFF
--- a/Test/TestsDatabase/OrderDatabaseTests.cs
+++ b/Test/TestsDatabase/OrderDatabaseTests.cs
@@ -208,7 +208,7 @@ namespace Test.TestsDatabase
         [InlineData(6)]
         public void OrdersCanBeWrittenToDatabaseWithTheoryWithHelper(int value)
         {
-            using (var context = new ContextHelper().ContextHelper2())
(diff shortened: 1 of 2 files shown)


srkirkland:
@jSylvestre yea the problem is that you want to close it, so I think it makes more sense to implement iDisposible and then put it in a using statement.

Test/TestsDatabase/OrderDatabaseTests.cs
```diff
@@ -199,5 +199,42 @@ public void OrdersCanBeWrittenToDatabaseWithTheory(int value)
...
+
+                    var order = CreateValidEntities.Order(i + 1);
+                    order.Creator = contextHelper.Context.Users.FirstOrDefault();
+                    contextHelper.Context.Orders.Add(order);
+                }
+
+
+                contextHelper.Context.SaveChanges();
+
+                var updatedOrders = contextHelper.Context.Orders.ToList();
```
srkirkland: @jSylvestre would be much more efficient to just do a count() here at the DB level instead of grabbing back the whole list and counting the elements

PR #20, opened 2017-05-02
Jcs/cache

jSylvestre:


Anlab.Mvc/Startup.cs
```diff
@@ -68,6 +68,7 @@ public void ConfigureServices(IServiceCollection services)
             services.AddTransient<IEmailSender, AuthMessageSender>();
             services.AddTransient<ISmsSender, AuthMessageSender>();
             services.AddTransient<IDirectorySearchService, DirectorySearchService>();
+            services.AddMemoryCache();
```
srkirkland: @jSylvestre you should add this higher, above AddMvc

PR #23, opened 2017-05-02
If User Name isn't set, assign the email to it.

jSylvestre:


Anlab.Mvc/Controllers/AccountController.cs
```diff
@@ -225,6 +225,10 @@ public IActionResult ExternalLogin(string provider, string returnUrl = null)
                     return View("ExternalLoginFailure");
                 }
                 var user = new User { UserName = model.Email, Email = model.Email, FirstName = model.FirstName, LastName = model.LastName, Name = model.Name };
+                if (string.IsNullOrWhiteSpace(user.Name))
```
srkirkland: @jSylvestre probably better to do in the constructor of user.cs

PR #30, opened 2017-05-04
Select Sample Type and filter list based on choice

jSylvestre:
Just playing with it, probably don't want to merge.


Anlab.Mvc/Client/components/SampleTypeSelection.tsx
```diff
@@ -0,0 +1,28 @@
+﻿import * as React from 'react';
+import { RadioGroup, RadioButton } from 'react-toolbox/lib/radio';
+
+export interface ISampleType {
+    sampleType: string;    
+}
+
+interface ISampleTypeProps {
+    sampleType: ISampleType;
```
srkirkland: @jSylvestre just make this type `string`

PR #32, opened 2017-05-04
Jcs/more fake data

jSylvestre:


srkirkland:
@jSylvestre looks good.  Note the only reason we would need someone's login in the initDB is if they have something attached to it, like an order.  Otherwise Cal would just get an account created for him when he logs in

PR #36, opened 2017-05-05
Use toFixed to format currency

jSylvestre:


jSylvestre:
But this would probably be better:
https://www.npmjs.com/package/react-number-format

PR #38, opened 2017-05-08
Whole numbers greater than zero

jSylvestre:


Anlab.Mvc/Client/components/numberInput/numberInput.tsx
```diff
@@ -29,7 +29,13 @@ export class NumberInput extends React.Component<INumberInputProps, INumberInput
         return value ? String(value) : '';
     }
     onChange = (v: string) => {
-        const error = isNaN(Number(v)) ? 'Must be a positive number' : null;
+        let error = isNaN(Number(v)) ? 'Must be a positive number' : null;
+        if (error == null) {
+            v = Number(v).toFixed(0).toString();
```
srkirkland: @jSylvestre try not to ever reassign a parameter. Best option would be do a `let value = v;` at the top and then change that if necessary.

Anlab.Mvc/Client/components/numberInput/numberInput.tsx
```diff
@@ -29,7 +29,13 @@ export class NumberInput extends React.Component<INumberInputProps, INumberInput
         return value ? String(value) : '';
     }
     onChange = (v: string) => {
-        const error = isNaN(Number(v)) ? 'Must be a positive number' : null;
+        let error = isNaN(Number(v)) ? 'Must be a positive number' : null;
```
srkirkland: @jSylvestre I'd like it better if the logic branched off of `if (isNan(value))` and then went two paths depending on if it was a valid number or not, instead of reusing the error string like this.

PR #39, opened 2017-05-08
Jcs/tool tip

jSylvestre:


srkirkland:
@jSylvestre there is already a tooltip component in the existing react-toolbox we are using http://react-toolbox.com/#/components/tooltip.  Also, bootstrap has a tooltip which we could use.  Try both of those before adding this new library and only if they don't work we can go this way

PR #40, opened 2017-05-08
Go to react-toolbox table

srkirkland:


jSylvestre:
@srkirkland FYI, this is ok, just wanted to make sure it is expected.
If you select a soil test, switch to water, then switch back to soil, soil test is still selected.
If you do the same thing except select a water test, then it clears out the soil test selection.

srkirkland:
not as useful as raw table

PR #51, opened 2017-05-10
First pass for details/confirmation page.

jSylvestre:


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -99,6 +106,19 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
...
+            if (response.success === true) {
+                var redirectId = response.id;
+                window.location.replace("/Order/Confirmation/" + redirectId);
+            } else {
+                alert("There was a problem saving your order.");
+                that.setState({ ...this.state, isSubmitting: false });
+            }
+            
+            }).error(function () {
+            that.setState({ ...this.state, isSubmitting: false });
```
jSylvestre: @srkirkland If I change "that" to "this", the state here doesn't get updated.

srkirkland: @jSylvestre but what about the state on line 115?  It's the exact same usage.

PR #57, opened 2017-05-22
Delete Order.

jSylvestre:


Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -233,6 +233,36 @@ public IActionResult Create()
             return RedirectToAction("Index");
 
         }
+
+        [HttpPost]
+        public async Task<IActionResult> Delete(int id)
+        {
+            var order = await _context.Orders.Include(i => i.Creator).SingleOrDefaultAsync(o => o.Id == id);
```
srkirkland: @jSylvestre no reason to need to include creator here, you aren't using it.

PR #61, opened 2017-05-22
Additional Fees (Grind, ForeignSoil, FilterWater)

jSylvestre:
And these amounts are different for internal

### DIFF
--- a/Anlab.Core/Models/OrderDetails.cs
+++ b/Anlab.Core/Models/OrderDetails.cs
@@ -29,6 +29,12 @@ namespace AnlabMvc.Models.Order
         public IList<string> AdditionalEmails { get; set; }
 
         public string Project { get; set; }
+
+        public string Grind { get; set; } //Soil and Plant
+
+        public string ForeignSoil { get; set; } //Soil
+
+        public string FilterWater { get; set; } //Water
     }
 
 }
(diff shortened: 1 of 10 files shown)


Anlab.Core/Models/OrderDetails.cs
```diff
@@ -29,6 +29,12 @@ public OrderDetails()
         public IList<string> AdditionalEmails { get; set; }
 
         public string Project { get; set; }
+
+        public string Grind { get; set; } //Soil and Plant
```
srkirkland: @jSylvestre I think these should be booleans, since it's really a yes/no question.

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -22,6 +26,9 @@ interface IOrderState {
     isValid: boolean;
     isSubmitting: boolean;
     additionalEmails: Array<string>;
+    grind: string;
+    foreignSoil: string;
+    filterWater: string;
```
srkirkland: @jSylvestre seems like these should all be booleans as well.

Anlab.Mvc/Client/components/Grind.tsx
```diff
@@ -0,0 +1,41 @@
...
+        this.props.handleChange('grind', grind);
+    }
+    render() {
+        if (!(this.props.sampleType === "Soil" || this.props.sampleType === "Plant")) {
+            return null;
+        }
+        return (
+            <div>
+                <label>Grind Samples?:</label>
+                <RadioGroup name='comic' value={this.props.grind} onChange={this.handleChange}>
```
srkirkland: @jSylvestre since there are only two options a checkbox might work better here, but go ahead and leave it for now and cal can decide.

PR #65, opened 2017-05-25
User profile updates

jSylvestre:


Anlab.Mvc/Controllers/ProfileController.cs
```diff
@@ -42,6 +42,8 @@ public ProfileController(ApplicationDbContext context)
                 userToUpdate.FirstName = user.FirstName;
                 userToUpdate.LastName = user.LastName;
                 userToUpdate.Name = user.Name;
+                userToUpdate.Phone = user.Phone;
+                userToUpdate.Account = user.Account;
```
srkirkland: @jSylvestre maybe about time for automapper huh 👍 :trollface:

PR #67, opened 2017-05-31
Wire up buttons.

jSylvestre:
PaymentSelection

### DIFF
--- a/Anlab.Mvc/Client/components/PaymentSelection.tsx
+++ b/Anlab.Mvc/Client/components/PaymentSelection.tsx
@@ -1,7 +1,5 @@
 ﻿import * as React from 'react';
-import { RadioGroup, RadioButton } from 'react-toolbox/lib/radio';
 import Input from 'react-toolbox/lib/input';
-import TooltipWrapper  from './tooltipWrapper/tooltipWrapper';
 
 export interface IPayment {
     clientType: string;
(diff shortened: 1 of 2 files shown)


Anlab.Mvc/Client/components/PaymentSelection.tsx
```diff
@@ -37,29 +34,23 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
...
-          <div className="anlab_form_style col-5"><h3>Credit Card</h3><p>It's amazing what you can do with a little love in your heart.</p></div>
-          <span className="dividing_span col-2 t-center align-middle">or</span>
-          <div className="anlab_form_style col-5 active-border active-text active-bg"><h3>UC Funds</h3><p>It's amazing what you can do with a little love in your heart</p></div></div>
-          </div>
-
-
+            <div className="form_wrap">
+                <h2 className="form_header">How will you pay for your order?</h2>
+                <div className="row">
+                    <div className={this.props.payment.clientType !== 'uc' ? "anlab_form_style col-5 active-border active-text active-bg" : "anlab_form_style col-5"} onClick={() => this.handleChange("other")}>
```
srkirkland: @jSylvestre pretty good, it'd be nice not to have to repeat the class names so much, maybe a lookup or const would do the trick, but for now I'm ok with it too since it'll probably change.

PR #69, opened 2017-06-01
Fix collapse

jSylvestre:
Removed extra closing div


jSylvestre:
@cydoval This collapses the other ones if a different one is selected.

PR #74, opened 2017-06-05
Explain why the button is disabled

jSylvestre:

### DIFF
--- a/Anlab.Mvc/Client/components/OrderForm.tsx
+++ b/Anlab.Mvc/Client/components/OrderForm.tsx
@@ -82,7 +82,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
         this.state = { ...initialState };
     }
     validate = () => {
-        const valid = this.state.quantity > 0 && !!this.state.project.trim();
+        const valid = this.state.quantity > 0 && this.state.quantity <= 100 && !!this.state.project.trim();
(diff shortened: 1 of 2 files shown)


jSylvestre:
image

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -82,7 +82,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
         this.state = { ...initialState };
     }
     validate = () => {
-        const valid = this.state.quantity > 0 && !!this.state.project.trim();
+        const valid = this.state.quantity > 0 && this.state.quantity <= 100 && !!this.state.project.trim();
```
srkirkland: @jSylvestre is there a way we can just ask the quantity box if it is valid, or have a this.state.quantityValid property?  That way we aren't tied in multiple places to this range limit

jSylvestre: @srkirkland I was asking @jpknoll about this. There are some frameworks we could import, but may not be worth it unless the form gets more complex.

srkirkland: @jSylvestre Yea not too interested in adding a forms lib, was just hoping there was a way of not duplicating the logic

PR #75, opened 2017-06-08
Jcs/admin role stuff

jSylvestre:


Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -2,24 +2,158 @@
...
+                return RedirectToAction("ListNonAdminUsers");
+            }
+            if (ModelState.IsValid)
+            {
+                userToUpdate.FirstName = user.FirstName;
+                userToUpdate.LastName = user.LastName;
+                userToUpdate.Name = user.Name;
+                userToUpdate.Phone = user.Phone;
+                userToUpdate.Account = user.Account;
+                userToUpdate.ClientId = user.ClientId;
```
srkirkland: @jSylvestre If you are just going to copy over vars like this, you don't need to restrict the binding on the method signature.  If you want to restrict the binding like you are doing, then there is no need to copy since only the proper fields are being bound.  I'm fine with it this way but there are different options, and maybe we can get automapper in one day too ;)

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -150,11 +155,16 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
             selected: filtered.filter(item => !!selectedTests[item.id])
         };
     }
-    onSubmit = () => {
-
+    onSubmit = () => {        
         if (this.state.isSubmitting) {
             return;
         }
+        var postUrl = '/order/save';
```
srkirkland: @jSylvestre If you can try to use let or const instead of var.

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -2,24 +2,228 @@
...
+            _userManager = userManager;
+            _roleManager = roleManager;
+            _orderService = orderService;
         }
-        public IActionResult Index()
+
+        [Authorize(Roles = RoleCodes.Admin)]
+        public async Task<IActionResult> Index()
+        {
+            var adminUsers = await _userManager.GetUsersInRoleAsync(RoleCodes.Admin);
```
srkirkland: @jSylvestre this is unused anywhere

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -2,24 +2,228 @@
...
         {
             _dbContext = dbContext;
+            _userManager = userManager;
+            _roleManager = roleManager;
+            _orderService = orderService;
         }
-        public IActionResult Index()
+
+        [Authorize(Roles = RoleCodes.Admin)]
+        public async Task<IActionResult> Index()
```
srkirkland: @jSylvestre you can replace this whole method with:

```
            var adminRole = await _roleManager.FindByNameAsync("Admin");
            var userRole = await _roleManager.FindByNameAsync("User");
            
            var usersRoles = _dbContext.Users.Select(u => new UserRolesModel
            {
                User = u,
                IsAdmin = u.Roles.Any(a => a.RoleId == adminRole.Id),
                IsUser = u.Roles.Any(a => a.RoleId == userRole.Id)
            }).ToList();
            
            return View(usersRoles);
```

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -2,24 +2,228 @@
...
+                }
+                usersRoles.Add(ur);
+            }           
+
+            return View(usersRoles);
+        }
+
+        public IActionResult ListNonAdminUsers()
+        {
+            var adminRole = _roleManager.Roles.Single(a => a.Name == RoleCodes.Admin).Id;
```
srkirkland: @jSylvestre I do like `var adminRole = await _roleManager.FindByNameAsync("Admin");` a bit better than this way, but both work fine.

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -2,24 +2,228 @@
...
+            var users = _dbContext.Users.Include(a => a.Roles).Where(w => w.Roles.All(x => x.RoleId != adminRole)).ToList();
```
srkirkland: @jSylvestre you don't need the include here at all, that will just result in returning more data than you need.

PR #80, opened 2017-06-15
Extra admin fields

jSylvestre:


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -57,6 +61,9 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
             isErrorActive: false,
             errorMessage: '',
             isAdmin: false,
+            status: '',
+            labComments: '',
+            adjustmentAmount: 0,
         };
```
srkirkland: @jSylvestre this initial state is starting to get really long.  Something to keep an eye on because it's only going to grow.  Maybe we can do more like what we are doing with payment and start nesting a bit.

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -170,7 +170,13 @@ public IActionResult ListUsersOrders(string id)
             {
                 var orderToUpdate = await _dbContext.Orders.SingleAsync(a => a.Id == model.OrderId.Value);
 
-                await _orderService.PopulateOrder(model, orderToUpdate);
+                await _orderService.PopulateOrder(model, orderToUpdate, true);
```
srkirkland: @jSylvestre don't love the mystery bool isAdmin, maybe it's better to just use User.IsInRole or do an overload PopulateLabOrder().

jSylvestre: @srkirkland That is just to tell me where the service is called from. I don't want any of those fields updated from the client side. But I'll think about it.

PR #82, opened 2017-06-15
Start email queue, move DB context to core project

srkirkland:
@jSylvestre still has a ways to go to send email (though it does work), but I wanted to get this merged before too long because of the big changes to the DB context location


Anlab.Core/Domain/MailMessage.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            SentAt = DateTime.UtcNow;
+        }
+
+        public int Id { get; set; }
+
+        [Required]
+        public string SendTo { get; set; }
+
+        public bool? Sent { get; set; }
```
jSylvestre: We usually use pending. But this is fine.

srkirkland: @jSylvestre my thought with sent was null means not sent, false means permanently not sent (error), and true means successful sent.

srkirkland:
@jSylvestre yea strange since i didn't change the order controller at all.  Guess it was a tabs vs. spaces or line endings thing.

PR #96, opened 2017-06-21
Jcs/js tests20170621

jSylvestre:


Anlab.Mvc/Client/components/AdditionalEmails.tsx
```diff
@@ -44,9 +50,9 @@ export class AdditionalEmails extends React.Component<IAdditionalEmailsProps, IA
             return null;
         }
 
-        const emails = this.props.addedEmails.map(item => {
+        const emails = this.props.addedEmails.map((item, index) => {
```
srkirkland: @jSylvestre you aren't using index and don't need it here

PR #97, opened 2017-06-22
Jcs/js tests20170622

jSylvestre:


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -236,7 +236,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
                     <AdditionalEmails addedEmails={additionalEmails} onEmailAdded={this.onEmailAdded} onDeleteEmail={this.onDeleteEmail}/>
                     <Project project={project} handleChange={this.handleChange} />
                     <AdditionalInfo additionalInfo={additionalInfo} handleChange={this.handleChange} />
-                    <LabFields isAdmin={isAdmin} labComments={labComments} adjustmentAmount={adjustmentAmount} handleChange={this.handleChange} />
+                    <LabFields isFromLab={isFromLab} labComments={labComments} adjustmentAmount={adjustmentAmount} handleChange={this.handleChange} />
```
srkirkland: @jSylvestre I'd rather you just now show this at all if they are not from the lab, instead of passing isFromLab in and then just returning null inside that component.

jSylvestre: @srkirkland How do I put a conditional in the orderForm render?

PR #99, opened 2017-06-29
LH/mail queue - make the mail queue look good 

laholstege:


Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+
+        </tr>
+    </thead>
+    <tbody style="vertical-align: top;">
+    @foreach (var message in Model)
+    {
+        <tr>
+            <td>
+
+                @if(message.Sent.HasValue && message.Sent == true)
```
srkirkland: @laholstege `message.Sent == true` is redundant, you don't need the == true part

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+                }
+                else
+                {
+                    <p>No</p>
+                }
+            </td>
+            <td>@message.SendTo
+                
+            </td>
+            <td data-search="@message.Subject">
```
srkirkland: 👍

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+                @message.Subject<br />
+            <a data-toggle="collapse" href="#@message.Id" aria-expanded="true" aria-controls="@message.Id">
+                Expand Body
+            </a> |
+            <a href="ViewMessage/@message.Id" target="_blank">Open in new window</a>
```
srkirkland: @laholstege To make sure routing works right in all cases you should use a tag helper here, something like this `<a asp-   asp-action="ViewMessage" asp-route-id="@message.Id"> ... </a>`

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
+﻿@model List<Anlab.Core.Domain.MailMessage>
+@{
+    ViewData["Title"] = "Mail Queue";
+}
+<div id="collapsebutton" class="nodisp expandcollapse btn btn-small btn-success no-print" style="margin: 10px 0px;"><i class="glyphicon glyphicon-minus"></i> Collapse All</div>
+<div id="expandbutton" class="disp expandcollapse btn btn-small btn-success no-print" style="margin: 10px 0px;"><i class="glyphicon glyphicon-plus"></i> Expand All</div>
+<div  style="padding-bottom:120px;">
```
jpknoll: Move this style to a class.

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+    <thead>
+        <tr>
+            <th>Sent:</th>
+            <th>To:</th>
+            <th>Subject:</th>
+            <th>Sent At:</th>
+
+        </tr>
+    </thead>
+    <tbody style="vertical-align: top;">
```
jpknoll: Move this style to a class.

Anlab.Mvc/Views/Admin/MailQueue.cshtml
jpknoll: You don't need to check for == true:

`@if (message.Sent.HasValue && message.Sent)`

Also, add the whitespace after if and (

jpknoll: You could probably use an Elvis operator here too.

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+    @foreach (var message in Model)
+    {
+        <tr>
+            <td>
```
jpknoll: You might turn this into a tertiary statement.

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+                    <a data-toggle="collapse" href="#@message.Id" aria-expanded="true" aria-controls="@message.Id">
+                        Collapse Body
+                    </a> |
+                    <a href="ViewMessage/@message.Id" target="_blank">Open in new window</a>
+                
+            </div>
+        </td>
+            <td>
+
+            @if (message.Sent.HasValue && message.Sent==true)
```
jpknoll: Use Elvis operator

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+                    },
+                    {
+                        width: "55%", "targets":[2]
+
+                    }
+                ]
+            });
+            $("#collapsebutton").hide();
+            $("#expandbutton").click(function () {
+                $('div.panel-collapse').addClass('in').css("height", "");
```
srkirkland: @laholstege what is the css height thing doing here?  Seems like you add it when expanding but don't remove it when collapsing.  Is it a bug in the collapse library you are trying to get around?

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+
+    </tbody>
+</table>
+    </div>
+
+@section AdditionalStyles
+{
+    <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.10.13/css/jquery.dataTables.min.css" />
+    <style>
+        div.panel-body tr tr{
```
jpknoll: Did you mean tr td?

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,125 @@
...
+{
+    <script type="text/javascript" src="https://cdn.datatables.net/1.10.13/js/jquery.dataTables.min.js"></script>
+    <script type="text/javascript">
+        $(function() {
+            $("#table").dataTable({
+                "order": [[0, "asc"],[3,"desc"]],
+                "columnDefs": [
+                    {
+                        "type": "ordering", "targets": [0, 1, 2, 3]
+
```
jpknoll: Remove these extra newlines

Anlab.Mvc/Views/Admin/MailQueue.cshtml
jpknoll: I think the collapse function supports a .collapse() call.

http://getbootstrap.com/javascript/#via-javascript-3

Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -0,0 +1,132 @@
...
+            <th>Status:</th>
+            <th>To:</th>
+            <th>Subject:</th>
+            <th>Sent At:</th>
+        </tr>
+    </thead>
+    <tbody>
+    @foreach (var message in Model)
+    {
+        <tr>
```
srkirkland: @laholstege no reason to change but in the future you can actually just create a variable for something like this.  `var sentStatus = message.Sent.HasValue ? "Sent" : "Failed"` and then in the actual <td> you can just do `@sentStatus`.  Have to put in an extra check but the idea is to get the logic into c# instead of doing a bunch of <p> tags.  Just an FYI you can keep this as-is

PR #102, opened 2017-07-05
Lh/receive Mail Message + Confirmation Page

laholstege:
adds confirmation page for receiving order, and adds mail message to mail queue once confirmed


Anlab.Mvc/Views/Lab/Confirmation.cshtml
```diff
@@ -0,0 +1,38 @@
+﻿@using Humanizer
+@model AnlabMvc.Models.Order.OrderReviewModel
+
+@{
+    ViewData["Title"] = "Confirmation";
+}
+
+<h2>Confirmation</h2>
```
jSylvestre: @laholstege Don't need the h2 here because it is using the title. Feel free to remove it from the order confirmation page too.

PR #105, opened 2017-07-06
More fields

jSylvestre:


Anlab.Core/Domain/MailMessage.cs
```diff
@@ -28,5 +28,9 @@ public MailMessage()
         public DateTime? SentAt { get; set; }
 
         public DateTime CreatedAt { get; set; }
+
+        public Order Order { get; set; }
```
srkirkland: @jSylvestre I didn't do this on purpose, I'm not sure we'll want to have every message associated with an order.  What if we want to send someone a message about their clientID or payment information?

jSylvestre: @srkirkland Just leave it null?

srkirkland: @jSylvestre sure we could, just don't see how any of this is worth the time/effort.  just trying to send an email here

PR #108, opened 2017-07-18
Prices from service.

jSylvestre:
First pass

### DIFF
--- a/Anlab.Core/Models/TestItemPrices.cs
+++ b/Anlab.Core/Models/TestItemPrices.cs
@@ -14,6 +14,8 @@ namespace Anlab.Core.Models
         public decimal SetupCost { get; set; }
         public string Name { get; set; }
         public int Multiplier { get; set; }
+
+        public decimal SetupPrice => SetupCost * Multiplier;
     }
 
 }
--- a/Anlab.Core/Models/TestItemPrices.cs
+++ b/Anlab.Core/Models/TestItemPrices.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Anlab.Core.Models
+{
+    public class TestItemPrices
+    {
+        [Required]
+        [StringLength(128)]
+        public string Code { get; set; }
+
+        public decimal Cost { get; set; }
+
+        public decimal SetupCost { get; set; }
+
+
+    }
+}
--- a/Anlab.Core/Models/TestItemPrices.cs
+++ b/Anlab.Core/Models/TestItemPrices.cs
@@ -2,19 +2,18 @@
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Text;
+using Anlab.Core.Domain;
 
 namespace Anlab.Core.Models
 {
+
     public class TestItemPrices
     {
-        [Required]
-        [StringLength(128)]
         public string Code { get; set; }
-
         public decimal Cost { get; set; }
-
         public decimal SetupCost { get; set; }
-
-
+        public string Name { get; set; }
+        public int Multiplier { get; set; }
     }
+
 }
--- a/Anlab.Mvc/Client/components/Summary.tsx
+++ b/Anlab.Mvc/Client/components/Summary.tsx
@@ -31,7 +31,7 @@ export class Summary extends React.Component<ISummaryProps, any> {
             const price = this.props.payment.clientType === 'uc' ? item.internalCost : item.externalCost;
             const perTest = price * this.props.quantity;
 
-            return prev + perTest + item.setupCost;
+            return prev + perTest + (this.props.payment.clientType === 'uc' ? item.internalSetupCost : item.externalSetupCost);
         }, 0);
 
         return total + grindTotal + foreignSoilTotal + filterWaterTotal + this.props.adjustmentAmount;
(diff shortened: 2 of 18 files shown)


Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -34,7 +37,28 @@ public OrderController(ApplicationDbContext context, IOrderService orderService,
...
+                    FeeSchedule = i.FeeSchedule,
+                    Group = i.Group,
+                    GroupType = i.GroupType,
+                    Id = i.Id,
+                    InternalCost = Math.Ceiling(p.Cost),
+                    SetupCost = Math.Ceiling(p.SetupCost),
+                    Multiplier = i.Multiplier,
+                    Multiplies = i.Multiplies,
+                    Notes = i.Notes,
+                    Public = i.Public}).ToList();
```
srkirkland: @jSylvestre TestItem needs

Code (assignable primary key Id)
Category
Group
Description
Comment

Anlab.Core/Models/TestItemPrices.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Text;
+
+namespace Anlab.Core.Models
+{
+    public class TestItemPrices
+    {
+        [Required]
+        [StringLength(128)]
```
srkirkland: @jSylvestre Remove

Anlab.Core/Models/TestItemPrices.cs
```diff
@@ -0,0 +1,20 @@
...
+        public string Code { get; set; }
+
+        public decimal Cost { get; set; }
+
+        public decimal SetupCost { get; set; }
+
```
srkirkland: @jSylvestre add Name, Multiplier (Work Effort Code?)

Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -34,7 +37,28 @@ public OrderController(ApplicationDbContext context, IOrderService orderService,
 
         public IActionResult Create()
         {
-            var model = new OrderEditModel { TestItems = _context.TestItems.AsNoTracking().ToArray() };
+            var prices = _itemPriceService.GetPrices();
+            var items = _context.TestItems.AsNoTracking().ToList();
+
+            var joined = (from i in items
+                join p in prices on i.Code equals p.Code
+                select new TestItem{Analysis = i.Analysis,
```
srkirkland: @jSylvestre New model here, TestItemModel or similar, that has these additional calculatable fields.

Anlab.Core/Models/TestItemPrices.cs
srkirkland: @jSylvestre also need IsActive

Anlab.Core/Models/TestItemPrices.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Anlab.Core.Models
+{
+
+    public class TestItemPrices
+    {
+        public string Code { get; set; }
+        public decimal Cost { get; set; }
+        public decimal SetupCost { get; set; }
+        public string Name { get; set; }
+        public int Multiplier { get; set; }
```
srkirkland: @jSylvestre As notes -- Setup Price is "WorkUnit" (aka multiplier) times SP-SetupPrice (aka SetupCost).  Probably worth adding some comments, and since we won't deal with either of those fields alone it could be worth adding a getter to determine setup price.

Anlab.Mvc/Client/components/Summary.tsx
```diff
@@ -22,56 +19,32 @@ interface ISummaryProps {
...
-        const foreignSoilTotal = this.foreignSoilCost() * this.props.quantity;
-        const filterWaterTotal = this.waterFilterCost() * this.props.quantity;
 
         const total = this.props.testItems.reduce((prev, item) => {
             // total for current item
             const price = this.props.payment.clientType === 'uc' ? item.internalCost : item.externalCost;
             const perTest = price * this.props.quantity;
 
-            return prev + perTest + item.setupCost;
+            return prev + perTest + (this.props.payment.clientType === 'uc' ? item.internalSetupCost : item.externalSetupCost);
```
srkirkland: @jSylvestre setup cost is not the same for internal/external so we'll still need to figure this out

Anlab.Mvc/Client/components/Summary.tsx
```diff
@@ -81,7 +54,7 @@ export class Summary extends React.Component<ISummaryProps, any> {
     }
 
     _renderAdditionalFees = () => {
-        if (!(this.props.grind || this.props.foreignSoil || this.props.filterWater || this.props.adjustmentAmount !== 0)) {
+        if (!(this.props.adjustmentAmount !== 0)) {
```
srkirkland: @jSylvestre clean this double negative up

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -16,22 +17,72 @@ public interface IOrderService
...
+        public OrderService(ApplicationDbContext context, ITestItemPriceService itemPriceService, IOptions<AppSettings> appSettings)
         {
             _context = context;
+            _itemPriceService = itemPriceService;
+            _appSettings = appSettings.Value;
+        }
+
+        public List<TestItemModel> PopulateTestItemModel()
+        {
+            var prices = _itemPriceService.GetPrices();
```
srkirkland: @jSylvestre needs to change in the future to an async so might as well do it now and just return Task.Result() so we don't have a massive change of method signatures later on

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -16,22 +17,72 @@ public interface IOrderService
...
+                    Analysis = i.Analysis,
+                    Category = i.Category,
+                    Code = i.Code,
+                    ExternalCost = Math.Ceiling(p.Cost * _appSettings.NonUcRate),
+                    Group = i.Group,
+                    Id = i.Id,
+                    InternalCost = Math.Ceiling(p.Cost),
+                    ExternalSetupCost = Math.Ceiling(p.SetupCost * _appSettings.NonUcRate),
+                    InternalSetupCost = Math.Ceiling(p.SetupCost)
+                }).ToList();
```
srkirkland: @jSylvestre this exact code is duplicated, remember DRY.  I think in this case a good way to go would be to have TestItemModel have a ctor() which takes in an item and a price and constructs this object.  Otherwise we'll have 2 places doing the same pricing calculations, which is asking for trouble

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -16,22 +17,72 @@ public interface IOrderService
...
+                {
+                    Analysis = i.Analysis,
+                    Category = i.Category,
+                    Code = i.Code,
+                    ExternalCost = Math.Ceiling(p.Cost * _appSettings.NonUcRate),
+                    Group = i.Group,
+                    Id = i.Id,
+                    InternalCost = Math.Ceiling(p.Cost),
+                    ExternalSetupCost = Math.Ceiling(p.SetupCost * _appSettings.NonUcRate),
+                    InternalSetupCost = Math.Ceiling(p.SetupCost)
```
srkirkland: @jSylvestre for setup cost I think you also need to multiply by the workUnit/multiplier

PR #109, opened 2017-07-20
Jcs/access anlab db

jSylvestre:
Note: Needs updated user secrets


Anlab.Mvc/AnlabMvc.csproj
```diff
@@ -55,6 +56,19 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="Helpers\QueryResource.Designer.cs">
```
srkirkland: @jSylvestre usually we put these in a Resources folder, Helpers folder is usually for stuff like dateTime formatting and string extensions

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -25,13 +28,15 @@ public class OrderService : IOrderService
...
         private readonly AppSettings _appSettings;
 
-        public OrderService(ApplicationDbContext context, ITestItemPriceService itemPriceService, IOptions<AppSettings> appSettings)
+        public OrderService(ApplicationDbContext context, ITestItemPriceService itemPriceService, IOptions<AppSettings> appSettings, IOptions<ConnectionSettings> connectionSettings)
         {
             _context = context;
-            _itemPriceService = itemPriceService;
+            _itemPriceService = itemPriceService;            
             _appSettings = appSettings.Value;
+            _connectionSettings = connectionSettings.Value; //Will need to push to Anlab.
```
srkirkland: @jSylvestre why do this, doesn't seem to be used anywhere

jSylvestre: @srkirkland I'm assuming that when we push data to anlab, we will need the connection then. I can remove it until we actually do that if you want.

PR #117, opened 2017-07-26
Jcs/pull tests done

jSylvestre:


Anlab.Core/Domain/Order.cs
```diff
@@ -38,6 +38,8 @@ public class Order :IDatedEntity
 
         public string Status { get; set; }
 
+        public string OrderRequest { get; set; }
```
srkirkland: @jSylvestre is this like request number?

Anlab.Mvc/Controllers/HomeController.cs
```diff
@@ -1,9 +1,19 @@
...
+//using AnlabMvc.Services;
+using Microsoft.AspNetCore.Mvc;
 
 namespace AnlabMvc.Controllers
 {
     public class HomeController : ApplicationController
     {
+        //private readonly ILabworksService _labworksService;
+
+        //public HomeController(ILabworksService labworksService)
```
srkirkland: @jSylvestre dump

Anlab.Mvc/Services/DbInitializationService.cs
```diff
@@ -49,6 +49,33 @@ public DbInitializationService(ApplicationDbContext context, UserManager<User> u
                 "AL-KCL",
                 "AL-OX",
                 "AL-PY",
+                "-BCL-P-IC",
```
srkirkland: @jSylvestre i know this is just test data so w/e, but I think the ones that start with dashes are not valid codes, they are like for QC or tests or something

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -121,6 +123,43 @@ public void PopulateOrderWithLabDetails(OrderSaveModel model, Order orderToUpdat
...
+            var tests = await PopulateSelectedTestsItemModel(testIds);
+
+            var orderDetails = orderToUpdate.GetOrderDetails();
+            var calcualtedTests = new List<TestDetails>();
+            foreach (var test in tests)
+            {
+                var cost = orderDetails.Payment.IsInternalClient ? test.InternalCost : test.ExternalCost;
+                var costAndQuantity = cost * orderDetails.Quantity;
+
+                calcualtedTests.Add(new TestDetails
```
srkirkland: @jSylvestre this seems like a good method to extract and make public, i.e. create TestDetails given a test and an order details.  I think it comes up somewhere else too.

Anlab.Mvc/Views/Lab/UpdateFromCompletedTests.cshtml
```diff
@@ -0,0 +1,35 @@
...
+</div>
+
+@section AdditionalStyles
+{
+    <link rel="stylesheet" type="text/css" href="https://cdn.datatables.net/1.10.13/css/jquery.dataTables.min.css" />
+}
+
+@section Scripts
+{
+    <script type="text/javascript" src="https://cdn.datatables.net/1.10.13/js/jquery.dataTables.min.js"></script>
```
srkirkland: @jSylvestre has nothing to do with this PR but it reminded me -- go ahead and create a set of partials for datatables, so you can do like `@Html.Partial("_DataTablesScripts")`.  Or if you have another idea that's good too, my goal is not to repeat the CDN path more than once, I'd hate to upgrade and have to change in each file.

Anlab.Core/Domain/Order.cs
jSylvestre: @srkirkland Yes. I can rename it if you want.

Anlab.Mvc/Services/DbInitializationService.cs
jSylvestre: @srkirkland Lets me see what comes back. For real data, we wouldn't include any of those in the DB so they would just get filtered out.

Anlab.Mvc/Services/IOrderService.cs
jSylvestre: @srkirkland Yeah. There are enough differences that I want to think about it some more.

Anlab.Mvc/Views/Lab/UpdateFromCompletedTests.cshtml
jSylvestre: @srkirkland K. This was just a copy and past of the confirmation page with a couple minor changes.

PR #121, opened 2017-07-27
Jcs/error if returned tests not found

jSylvestre:
In this example, 1 was not found and the code is actually ()

image


jSylvestre:
image

srkirkland:
@jSylvestre what the heck code is ()?

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -108,17 +108,26 @@ private List<TestItemModel> GetJoined(IList<TestItemPrices> prices, List<TestIte
         /// </summary>
         /// <param name="orderToUpdate"></param>
         /// <returns></returns>
-        public async Task OverwiteOrderWithTestsCompleted(Order orderToUpdate)
+        public async Task<string> OverwiteOrderWithTestsCompleted(Order orderToUpdate)
```
srkirkland: @jSylvestre I don't like the idea that we just have to know the return value is a string containing errors.  How about creating a return object like OverwriteOrderResult and it can have an errors property which is a string array.   Then you can check if `result.Errors.length > 0` instead of having to do string logic.

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -123,8 +124,9 @@ private List<TestItemModel> GetJoined(IList<TestItemPrices> prices, List<TestIte
                 //Oh No!!! tests were returned that we don't know about
                 var foundCodes = _context.TestItems.Where(a => testIds.Contains(a.Id)).Select(s => s.Id).Distinct().ToList();
                 var missingCodes = testCodes.Except(foundCodes).ToList();
+                rtValue.Errors.Add(string.Format("Error. Unable to continue. The following codes were not found locally: {0}", string.Join(",", missingCodes)));
 
-                return string.Format("Error. Unable to continue. The following codes were not found locally: {0}", string.Join(",", missingCodes));
+                return rtValue;
```
jSylvestre: @srkirkland Better?

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -108,28 +107,38 @@ private List<TestItemModel> GetJoined(IList<TestItemPrices> prices, List<TestIte
...
-            var testIds = _context.TestItems.Where(a => testCodes.Contains(a.Code)).Select(s => s.Id).ToArray(); //TODO: Currently if the test code doesn't exist in our DB we ignore it, but this could cause issues if the test not found has a $ amount.
+            var testIds = _context.TestItems.Where(a => testCodes.Contains(a.Id)).Select(s => s.Id).ToArray(); 
             var tests = await PopulateSelectedTestsItemModel(testIds);
 
+            if (testCodes.Count != testIds.Length)
+            {
+                //Oh No!!! tests were returned that we don't know about
+                var foundCodes = _context.TestItems.Where(a => testIds.Contains(a.Id)).Select(s => s.Id).Distinct().ToList();
+                var missingCodes = testCodes.Except(foundCodes).ToList();
+                rtValue.Errors.Add(string.Format("Error. Unable to continue. The following codes were not found locally: {0}", string.Join(",", missingCodes)));
```
srkirkland: @jSylvestre I don't think this method which should overwrite orders should need to be formatting error strings.  I'd do `rtValue.MissingCodes = testCodes.Except(foundCodes).ToArray()` and then set rtValue.Success = true/false depending on status, and let the calling code decide how to handle missing codes and format the values on the caller side.

Anlab.Mvc/Services/IOrderService.cs
srkirkland: @jSylvestre better!

Issue #116, opened 2017-07-25
Column for test item icon


jSylvestre:
When a test item has notes associated with it, have an icon in the test items table to indicate that.
Maybe a null able int that we can have associated  with a font awesome icon

PR #128, opened 2017-08-02
added icon for notes of test items

laholstege:


Anlab.Mvc/Client/components/TestList.tsx
```diff
@@ -68,9 +68,12 @@ export class TestList extends React.Component<ITestListProps, ITestListState> {
                         <td>
                             <Checkbox checked={selected} onChange={e => this.onSelection(item, e)} />
                         </td>
-                        <td><div className="analysisTooltip" data-toggle="tooltip" title={item.notes}>{item.analysis}</div></td>
+                        <td>{item.analysis}</td>
                         <td>{item.id}</td>
                         <td><NumberFormat value={priceDisplay} displayType={'text'} thousandSeparator={true} decimalPrecision={true} prefix={'$'} /></td>
+                        <td width="5%">
+                            {item.notes ? <i className="fa fa-info-circle" aria-hidden="true" id="analysisTooltip" data-toggle="tooltip" title={item.notes}></i> : ""}
```
srkirkland: @laholstege there shouldn't ever be two elements in the DOM with the same ID, so you'll want to go with className here like it was before

jSylvestre:
Issues #116

jSylvestre:
@srkirkland Do we want to display the notes tip on the confirmation pages too?

Anlab.Core/Domain/TestItem.cs
```diff
@@ -29,6 +29,7 @@ public class TestItem
         public bool Public { get; set; }
 
         public string Notes { get; set; }
+        public string NotesEncoded { get; set; }
```
srkirkland: @laholstege What do you think about not having NotesEncoded be a settable property and just have it only have a getter.  That getter would be implemented to automatically encode the notes property, that way you wouldn't have to worry about setting it in the db initializer and every other place you need to use it

Issue #127, opened 2017-08-02
[ValidateAntiForgeryToken]


jSylvestre:
Add where missing

PR #130, opened 2017-08-07
Jcs/minor updates

jSylvestre:


jSylvestre:
Closes issue #127

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -74,7 +74,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
 
             initialState.quantity = orderInfo.Quantity;
             initialState.additionalInfo = orderInfo.AdditionalInfo;
-            initialState.additionalEmails = orderInfo.AdditionalEmails;
+            initialState.additionalEmails = orderInfo.AdditionalEmails != null? orderInfo.AdditionalEmails : [];
```
srkirkland: @jSylvestre try to never use double equals in javascript, but also what I would do here is just guarantee on the server-side that you never get back a null array for additional emails, in the ctor() have it new up an empty array

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -97,6 +97,8 @@ public IActionResult EditUser(string id, [Bind("FirstName,LastName,Name,Phone,Ac
         }
 
         [Authorize(Roles = RoleCodes.Admin)]
+        [HttpPost]
+        [ValidateAntiForgeryToken]
```
srkirkland: @jSylvestre do you need this validate attribute since you put one on the application controller?

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -176,7 +176,8 @@ private static void CalculateTest(OrderDetails orderDetails, TestItemModel test,
             orderDetails.Total = orderDetails.SelectedTests.Sum(x=>x.Total);
 
             orderToUpdate.SaveDetails(orderDetails);
-            orderToUpdate.AdditionalEmails = string.Join(";", orderDetails.AdditionalEmails);
+
+            orderToUpdate.AdditionalEmails = orderDetails.AdditionalEmails != null ? string.Join(";", orderDetails.AdditionalEmails): null;
```
srkirkland: @jSylvestre another point where if you just defaulted to an empty array you wouldn't need the check.

Anlab.Mvc/Controllers/AdminController.cs
jSylvestre: @srkirkland No. I didn't remove the ones that were already there.

srkirkland: @jSylvestre ah, yea i wouldn't have noticed except the attribute was added in this PR.

PR #131, opened 2017-08-10
Show tips on Confirmation Page

jSylvestre:


jSylvestre:
@laholstege @srkirkland If we want the tips on the confirmation page, we could do it like this.

PR #133, opened 2017-08-15
Create a partial script for the shadow tooltip

jSylvestre:


Anlab.Mvc/Views/Shared/_ShowdownScriptsPartial.cshtml
```diff
@@ -0,0 +1,34 @@
+<environment names="Development">
```
srkirkland: @jSylvestre really no reason to put things in a different environment if they aren't different.  Plus at the most you'd want to put the CDN source in the environment tag and leave the actual javascript outside of it.

PR #137, opened 2017-08-16
Jcs/payments

jSylvestre:


Anlab.Mvc/Controllers/PaymentController.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+            if (order.CreatorId != CurrentUserId)
+            {
+                ErrorMessage = "You don't have access to this order.";
+                return RedirectToAction("Index", "Order");
+            }
+
+            if (order.Status != OrderStatusCodes.Complete)
+            {
+                ErrorMessage = "You can't edit an order that has been confirmed.";
```
srkirkland: @jSylvestre We might want a different status in here but for now completed is ok.  Afterwards we can move to a "Paid" status or similar.  Probably will require changing this error message though because the edit text doesn't make sense

Anlab.Mvc/Models/Configuration/CyberSourceSettings.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace AnlabMvc.Models.Configuration
+{
+    public class CyberSourceSettings
```
srkirkland: @jSylvestre I like how you made these separate settings

Anlab.Mvc/Services/DataSigningService.cs
```diff
@@ -0,0 +1,69 @@
...
+                return signature == Sign(BuildDataToSign(paramsArray), _cyberSourceSettings.SecretKey);
+            }
+            catch (Exception ex)
+            {
+                Log.Error(ex, ex.Message);
+                return false;
+            }
+        }
+
+        private static string Sign(string data, string secretKey)
```
srkirkland: @jSylvestre maybe give this a slightly different name, SignData() or something, because otherwise I think it's confusing that it has the same name as a public interface method on the same class

Anlab.Mvc/Startup.cs
```diff
@@ -86,6 +87,7 @@ public void ConfigureServices(IServiceCollection services)
             //services.AddTransient<ILabworksService, FakeLabworksService>(); //TODO: Replace with non fake one.
             services.AddTransient<ILabworksService, LabworksService>();
             services.AddTransient<IFileStorageService, FileStorageService>();
+            services.AddSingleton<IDataSigningService, DataSigningService>();
```
srkirkland: 😄

Anlab.Mvc/Services/DataSigningService.cs
jSylvestre: @srkirkland Basically the same as give. But an easy change.

Anlab.Core/Domain/PaymentEvent.cs
```diff
@@ -0,0 +1,34 @@
...
+            OccuredAt = DateTime.UtcNow;
+        }
+        [Key]
+        public string Transaction_Id { get; set; }
+
+        /// <summary>
+        /// Unique merchant-generated order reference or tracking number for each transaction.
+        /// (Order Id)
+        /// </summary>
+        public int Req_Reference_Number { get; set; }
```
srkirkland: @jSylvestre in addition to req request number, which is the orderID, we should really have a FK link to the Order itself, you know like `public Order Order {get;set;}`

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -264,5 +264,18 @@ public IActionResult ListUsersOrders(string id)
...
         }
 
+        [Authorize(Roles = RoleCodes.Admin)]
+        public async Task<JsonResult> JsonDetails(int id)
+        {
+            var order = await _dbContext.Orders.SingleOrDefaultAsync(o => o.Id == id);
+
+            if (order == null)
+            {
+                return null;//NotFound(id);
```
srkirkland: @jSylvestre why?  I know we still need to setup a not found page but it seems like notfound is the way to go

Anlab.Mvc/Controllers/PaymentController.cs
```diff
@@ -0,0 +1,337 @@
...
+
+            var responseValid = CheckResponse(response);
+            if (!responseValid.IsValid)
+            {
+                ErrorMessage = ErrorMessage = string.Format("Errors detected: {0}", string.Join(",", responseValid.Errors));
+                return RedirectToAction("Pay", new {id = order.Id});
+            }
+
+            //Should be good,   
+            if (order.Status == OrderStatusCodes.AwaitingPayment)
```
srkirkland: @jSylvestre seems weird that you are updating the order both in the payment receipt and in the provider notify.  I think you can do it in either/both w/ a check to make sure it isn't already paid, but half here and full in the provider notify is a strange state of affairs because now you can have a paid order w/o an associated payment event for a while (or forever if the provider notify doesn't work)

Anlab.Mvc/Controllers/PaymentController.cs
```diff
@@ -0,0 +1,337 @@
...
+                // check signature
+                var dictionary = Request.Form.ToDictionary(x => x.Key.ToString(), x => x.Value.ToString());
+                if (!_dataSigningService.Check(dictionary, response.Signature))
+                {
+                    Log.Error("Check Signature Failure");
+                    ErrorMessage = "An error has occurred. Payment not processed. If you experience further problems, contact us.";
+                    return RedirectToAction("Index", "Home");
+                }
+                //ViewBag.PaymentDictionary = dictionary; //Debugging. Remove when not needed
+                ProcessPaymentEvent(response, dictionary); //TODO: Do we want to try to write cancel events?
```
srkirkland: @jSylvestre if we do we'll have to be careful because we'll want to find the payment even associated w/ an order.  If there are multiple then it could get strange if we don't pay attention

Anlab.Mvc/Controllers/PaymentController.cs
```diff
@@ -0,0 +1,337 @@
...
+                    Log.Error("Order not found {0}", response.Req_Reference_Number);
+                }
+                else
+                {
+                    if (response.Decision == ReplyCodes.Accept)
+                    {
+                        if (order.Status == OrderStatusCodes.AwaitingPayment)
+                        {
+                            order.Status = OrderStatusCodes.Paid;
+                            _context.SaveChanges(true);
```
srkirkland: @jSylvestre I wonder if you'd want to do order.Payment = payment, and then a single save would save both the payment info as well as the updated order.  That seems simple and atomic and would make future querying easier.

Anlab.Mvc/Views/Lab/Confirmation.cshtml
```diff
@@ -2,7 +2,8 @@
 @model AnlabMvc.Models.Order.OrderReviewModel
 
 @{
-    ViewData["Title"] = "Confirmation";
+    ViewData["Title"] = "Confirmation";
+    ViewBag.OrderId = Model.Order.Id;
```
srkirkland: @jSylvestre why is this needed?  same thing w/ a few other files too

Anlab.Mvc/Views/Payment/Pay.cshtml
```diff
@@ -0,0 +1,53 @@
...
+{
+    @{await Html.RenderPartialAsync("_DataTableScriptsPartial"); }
+
+    <script type="text/javascript">
+        $(function() {
+            $("#table").dataTable();
+        });
+    </script>   
+
+    @{await Html.RenderPartialAsync("_ShowdownScriptsPartial"); }
```
srkirkland: @jSylvestre why does the receipt page need the showdown scripts?

Anlab.Mvc/Views/Shared/Layout.cshtml
```diff
@@ -115,11 +115,19 @@
...
-<footer>
-  <div class="footer_contents"><p class="discreet">&copy; The Regents of the University of California, Davis campus, 20017-@(DateTime.UtcNow.Date.ToString("yy")).</p>
-    <p class="discreet">Version: @VersionFinder.Version</p></div>
-
 
+<footer>
+    <div class="footer_contents">        
+        <p class="discreet">
+            &copy; The Regents of the University of California, Davis campus, 20017-@(DateTime.UtcNow.Date.ToString("yy")).
+        @if (User != null && User.IsInRole(RoleCodes.Admin) && ViewBag.OrderId != null)
```
srkirkland: @jSylvestre oh i see.  Sorry to say but let's not do this, I don't want to search user roles on ever page load, even public pages.  We can put the lab details button into the header or something if we want, or don't even put a link to it and we can just know the URL.  We're the only ones that are going to use it anyway.  But it shouldn't be on the layout page

Anlab.Core/Domain/PaymentEvent.cs
jSylvestre: @srkirkland This is info back from CyberSource, I wanted to capture everything then return back even bad stuff, so I didn't want FKs. Let me know if you still want that.

Anlab.Mvc/Controllers/PaymentController.cs
jSylvestre: @srkirkland I believe that was what @jpknoll suggested, unless I misunderstood.

jpknoll: @jSylvestre Yeah, early returns, unpack the else statements.

Anlab.Mvc/Controllers/PaymentController.cs
jSylvestre: @srkirkland The PaymentEvents contains all failures and cancels as well, so I \'m not sure if that would work.

Anlab.Mvc/Views/Lab/Confirmation.cshtml
jSylvestre: @srkirkland The ViewBag.OrderId tells the layout page to show the </> for the OrderDetails json call.

Anlab.Mvc/Views/Payment/Pay.cshtml
jSylvestre: @srkirkland I'm showing the Order details partial view. It shows the test item test notes.

Anlab.Mvc/Views/Shared/Layout.cshtml
jSylvestre: @srkirkland We do it anyway for the menu above. I'm ok with changing it though.

srkirkland: @jSylvestre  Cool yea we should just dump the link part

Anlab.Mvc/Controllers/PaymentController.cs
srkirkland: @jSylvestre but in this case you only make the order paid if you know you have a successful order, so i don't know why it matters, you'll only have the one success for any given order.

jSylvestre:
Payment Events for 1 order:

image

PR #138, opened 2017-08-16
Lh/remove lab component

laholstege:


Anlab.Mvc/Views/Lab/UpdateFromCompletedTests.cshtml
```diff
@@ -19,6 +19,12 @@
         {
             <div>Results File Identifier: @Model.Order.ResultsFileIdentifier</div>
         }
+        <span>Add Lab Comments</span>
+        <input type="text" class="form-control" name="labComments" />
```
jSylvestre: @laholstege Should be a text area I think so we can have multiple lines.

Anlab.Core/Models/OrderDetails.cs
```diff
@@ -23,6 +23,11 @@ public class OrderDetails
 
         public string LabComments { get; set; }
         public decimal AdjustmentAmount { get; set; }
+        public decimal GrandTotal { get
```
jSylvestre: @laholstege Lets double check with @srkirkland about this one.
I don't think it is needed because it is getting added in the PopulateOrderWithLabDetails method already. But this method would need to be called. It doesn't need to be called in the save method anymore because that is just for the react form.

srkirkland: @jSylvestre But isn't that method only populating the Total?  Once we have adjustment amount we need to add it in, and that's what the grand total is for.  Or am I missing what you are saying here?

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -201,7 +213,7 @@ public IActionResult ListUsersOrders(string id)
             return View(model);
         }
         [HttpPost]
-        public async Task<IActionResult> UpdateFromCompletedTests(int id, bool confirm, IFormFile uploadFile)
+        public async Task<IActionResult> UpdateFromCompletedTests(int id, bool confirm, IFormFile uploadFile, string labComments, decimal adjustmentAmount)
```
srkirkland: @laholstege Not sure how easy it would be but if you could put this into an UpdateFromCompletedTestsModel.cs or something and have the method only take the (id, model) as params and let model binding take care of the rest, that would be a lot cleaner.

Anlab.Mvc/Views/Lab/UpdateFromCompletedTests.cshtml
```diff
@@ -46,7 +52,18 @@
...
 
+    <script type="text/javascript">
+        $('#adjustmentAmount').on('keyup', function () {
+            var prevTotal = $('#prevTotal').val();
+            var adjustAmt = $(this).val();
+            if (adjustAmt)
+                var totalCost = parseFloat(prevTotal) + parseFloat(adjustAmt);
+            else
+                var totalCost = prevTotal;
+            $('.newTotal').html(totalCost);
```
srkirkland: @laholstege update that span to have an id="newTotal" or similar and use the ID here, just to make sure you are updating only the one proper element

Anlab.Core/Models/OrderDetails.cs
jSylvestre: @srkirkland I was just using the Total before. Didn't see a need to create a new field. I'm fine either way.

srkirkland: I think w/ the adjustment being there it's good to have separate values

laholstege: I think we added it so that we can show the previous price and the new adjusted one to the user

PR #140, opened 2017-08-17
Lh/results

laholstege:


Anlab.Mvc/Views/Results/Index.cshtml
```diff
@@ -0,0 +1,64 @@
+@using AnlabMvc.Models.Order
+@model IList<Order>
```
srkirkland: @laholstege you can dump this whole index page

Issue #141, opened 2017-08-22
Test Items Key


jSylvestre:
I think the key must be made up of the id (code) and the Category. Otherwise we will have issues with Grind and Wrap

jSylvestre:
No. Bad.bad.Bad...

PR #142, opened 2017-08-23
Jcs/test item changes20170823

jSylvestre:


jSylvestre:
Closes #141

jSylvestre:
Need to fix using one in an order

jSylvestre:
Don't do this way. Closing PR

PR #146, opened 2017-08-28
Lh/admin all orders

laholstege:


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -25,6 +26,8 @@ public class LabController : ApplicationController
         private readonly IOrderMessageService _orderMessageService;
         private readonly IFileStorageService _fileStorageService;
 
+        private readonly int _maxShownOrders = 1000;
```
srkirkland: @laholstege since it is statically set, make it a const or a static readonly.  Const probably makes the most sense

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -33,15 +36,27 @@ public LabController(ApplicationDbContext dbContext, IOrderService orderService,
...
 
-        public IActionResult OpenOrders()
+        public IActionResult Orders()
         {
-            //TODO: update this when we know status. Add filter?
-            var orders = _dbContext.Orders.Where(a => a.Status != OrderStatusCodes.Created && a.Status != OrderStatusCodes.Complete)
-                .Include(i => i.Creator).ToList();
+            var orders = _dbContext.Orders
+                .Where(a => a.Status != OrderStatusCodes.Created)
+                .Select(c => new Order
```
srkirkland: @laholstege 👍 for select statement

PR #147, opened 2017-08-28
calculate adjustment amount from final total

laholstege:


Anlab.Core/Models/OrderDetails.cs
```diff
@@ -23,7 +23,7 @@ public class OrderDetails
         [DataType(DataType.MultilineText)]
         public string LabComments { get; set; }
         public decimal AdjustmentAmount { get; set; }
-        public decimal GrandTotal { get
+        public decimal FinalTotal { get
```
jSylvestre: @laholstege I'm using the GrandTotal in my Payment PR. If it isn't too much of a problem, please change FinalTotal back to GrandTotal.

PR #148, opened 2017-08-29
Jcs/minor fixes20170829

jSylvestre:


Anlab.Mvc/Views/Shared/Layout.cshtml
```diff
@@ -119,7 +119,7 @@
 <footer>
     <div class="footer_contents">        
         <p class="discreet">
-            &copy; The Regents of the University of California, Davis campus, 20017-@(DateTime.UtcNow.Date.ToString("yy")).
+            &copy; The Regents of the University of California, Davis campus, 2017-@(DateTime.UtcNow.Date.ToString("yy")).
```
srkirkland: @jSylvestre Why?  Won't this do 2017-17?  What's what supposed to mean?  Do you just want to do Date.ToString("yyyy")?

jSylvestre: @srkirkland We do similar things in other projects. Copyright is year first created, then current year.

srkirkland: @jSylvestre oh so you want it to say 2017-17?  I guess that's ok?

PR #152, opened 2017-08-29
Lh/client id on react

laholstege:


Anlab.Mvc/Client/components/ClientId.tsx
```diff
@@ -0,0 +1,51 @@
...
+    handleChange: Function;
+    isFromLab: boolean;
+}
+
+interface IClientIdInputState {
+    internalValue: string;
+    error: string;
+}
+
+export class ClientId extends React.Component<IClientIdProps, IClientIdInputState> {
```
srkirkland: @laholstege yay for adding in state and prop interfaces

Anlab.Mvc/Client/components/ClientId.tsx
```diff
@@ -0,0 +1,51 @@
...
+        super(props);
+
+        this.state = {
+            internalValue: this.props.clientId,
+            error: null
+        };
+    }
+    validate = (v: string) => {
+        let error = null;
+        if (v.trim() === "") {
```
srkirkland: @laholstege this works but if v is null it would error (not sure that can happen).  Making the if statement `!v || v.trim() === ''` will make sure it isn't null or undefined.  If you just care about null or empty strings, if (!v) {} would work

Issue #113, opened 2017-07-25
Setup Cost


jSylvestre:
If this isn't in the DB somewhere, use a config setting.

jSylvestre:
Currently hard coded here:
https://github.com/ucdavis/Anlab/blob/master/Anlab.Core/Models/TestItemPrices.cs#L14

jSylvestre:
It will be in the database somewhere. We will need to grab it and probably cache it for a days or so.

jSylvestre:
#155

PR #154, opened 2017-08-30
Update the query to populate the SetupCost

jSylvestre:


jSylvestre:
closes issue #113

srkirkland:
@jSylvestre I don't see why we should show setup price at all in this test items controller.  The only place it is needed is on the order pages (create/details/..).

PR #156, opened 2017-08-30
Lh/processing fee

laholstege:


Anlab.Mvc/Client/components/specs/Summary.spec.tsx
```diff
@@ -21,7 +21,8 @@ describe('<Summary />', () => {
                                                 hideError={true}
                                                 isCreate={true}
                                                 onSubmit={null}
-                                                testItems={testItems} />);
```
laholstege: @jSylvestre I wasn't sure how to properly adjust the spec for this so I just added processingFee as null for this. Is that okay?

Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -39,8 +42,13 @@ public OrderController(ApplicationDbContext context, IOrderService orderService,
         public async Task<IActionResult> Create()
         {
             var joined = await  _orderService.PopulateTestItemModel();
+            var proc = await _labworksService.GetPrice("PROC");
```
srkirkland: @laholstege move this into a const or an app setting

Anlab.Core/Models/OrderDetails.cs
```diff
@@ -29,6 +29,8 @@ public class OrderDetails
             }
         }
         public string ClientId { get; set; }
+        public decimal InternalProcessingFee { get; set; }
+        public decimal ExternalProcessingFee { get; set; }
```
srkirkland: @laholstege I'd rather you dump this and just have one processingFee property.  It'll take a small amount of extra work up front, but just a little bit.  That part in react where you say `const processingFee = this.state.payment.clientType === 'uc' ? this.state.internalProcessingFee : this.state.externalProcessingFee; ` you'll have to change to do the math.Ceil(this.state.processingFee) logic, and of course you'll have to pass the appsettings.NonUcRate into the react view in order to do that.  But then every other place in here where you have two ProcessingFee props can go away

Anlab.Mvc/Client/components/specs/Summary.spec.tsx
jSylvestre: @laholstege Yes, that is fine. That is what I would have done.

Anlab.Mvc/Controllers/OrderController.cs
jSylvestre: I'd say Const.

PR #157, opened 2017-08-31
Jcs/nice errors

jSylvestre:


jSylvestre:
Closes issue #129

Anlab.Mvc/Controllers/HomeController.cs
```diff
@@ -14,9 +15,20 @@ public IActionResult SamplingAndPreparation()
...
-            return View();
+            throw new Exception("Test exception. If this was a real exception, you would need to run in circles, scream and shout.");
+        }
+
+        public IActionResult TestNotFound1()
+        {
+            return NotFound();
+        }
+
+        public IActionResult TestNotFound2()
```
srkirkland: @jSylvestre maybe dump this second one and just have TestNotFound take an optional string id param.  That way you could test both in one go and it'd be cleaner

PR #168, opened 2017-09-05
Lh/admin test item filtering

laholstege:


Anlab.Mvc/Views/TestItems/Index.cshtml
```diff
@@ -81,17 +88,35 @@
     @{ await Html.RenderPartialAsync("_DataTableScriptsPartial"); }
     <script type="text/javascript">
         $(function() {
-            $("#table").dataTable({
+            var table = $("#table").DataTable({
```
srkirkland: @laholstege does this work even w/ upper-cased DataTable()?  I would think it would need to be .dataTable().  But maybe they aliased it

Anlab.Mvc/Views/TestItems/Index.cshtml
```diff
@@ -81,17 +88,35 @@
...
+            });
         });
+
+        $.fn.dataTable.ext.search.push(
+            function (settings, data, dataIndex) {
+
+                var category = data[3];
+                var categoryFilter = $('#category').val();
+
+                if (categoryFilter == "All" || category.indexOf(categoryFilter) !== -1) {
```
srkirkland: @laholstege just to check, category (data[3]) is a string right?  If so is there a need to do string.indexOf instead of strict equality (i.e. `category === categoryFilter`)?

laholstege: @srkirkland category can be a string like "Soil Water" so it's just to make sure those get included

srkirkland: @laholstege ah makes sense.  Ok 🚢

PR #175, opened 2017-09-07
Lh/historic test prices

laholstege:


Anlab.Core/Domain/Order.cs
```diff
@@ -69,5 +70,22 @@ public void SaveDetails(OrderDetails details)
             JsonDetails = JsonConvert.SerializeObject(details);
         }
 
+        public IList<TestItemModel> GetTestDetails()
+        {
+            try
+            {
+                return JsonConvert.DeserializeObject<IList<TestItemModel>>(SavedTestDetails);
+            }
+            catch (JsonSerializationException)
```
srkirkland: @laholstege this catch does nothing except eat the stack trace, so just remove the whole try/catch

PR #177, opened 2017-09-07
Lh/status updates

laholstege:
update filters and mail messages based on new statuses


Anlab.Mvc/Services/OrderMessageService.cs
```diff
@@ -75,23 +74,32 @@ public OrderMessageService(ViewRenderService viewRenderService, IMailService mai
             _mailService.EnqueueMessage(message);
         }
 
-        public async Task EnqueueCompletedMessage(Order order)
+        public async Task EnqueueAcceptedMessage(Order order)
```
srkirkland: @laholstege i'm not sure if we are going to send an email for this case, but for now we can leave it in

Anlab.Mvc/Views/Order/Index.cshtml
```diff
@@ -62,27 +86,68 @@
 @section AdditionalStyles
 {
     @{ await Html.RenderPartialAsync("_DataTableStylePartial"); }
+    <link rel="stylesheet" type="text/css" href="https://code.jquery.com/ui/1.12.1/themes/base/jquery-ui.css" />
```
srkirkland: @laholstege We usually use the bootstrap datepicker instead of bringing in the whole jqueryUI library.  The change probably won't be too hard. https://github.com/uxsolutions/bootstrap-datepicker.  There's a CDN you can use to grab the min css and javascript for that project that they mention in the readme.

PR #181, opened 2017-09-11
use status codes on filters

laholstege:


Anlab.Mvc/Views/Lab/Orders.cshtml
```diff
@@ -17,11 +17,11 @@
...
-                    <option value="Confirmed">Confirmed</option>
-                    <option value="Received">Received</option>
-                    <option value="Finalized">Finalized</option>
-                    <option value="Accepted">Accepted</option>
-                    <option value="Complete">Complete</option>
+                    <option value="@OrderStatusCodes.Confirmed">@OrderStatusCodes.Confirmed</option>
+                    <option value="@OrderStatusCodes.Received">@OrderStatusCodes.Received</option>
+                    <option value="@OrderStatusCodes.Finalized">@OrderStatusCodes.Finalized</option>
+                    <option value="@OrderStatusCodes.Accepted">@OrderStatusCodes.Accepted</option>
+                    <option value="@OrderStatusCodes.Complete">@OrderStatusCodes.Complete</option>
```
srkirkland: @jSylvestre do you remember how we do this in GREAT?  I'm thinking make we throw the OrderStatusCodes into an array and then do a foreach?  Just want to future proof for the case where we add another status so we don't have to change this

Anlab.Mvc/Views/Order/Index.cshtml
```diff
@@ -22,12 +22,12 @@
...
-                    <option value="Received">Received</option>
-                    <option value="Finalized">Finalized</option>
-                    <option value="Accepted">Accepted</option>
-                    <option value="Complete">Complete</option>
+                    <option value="@OrderStatusCodes.Created">@OrderStatusCodes.Created</option>
+                    <option value="@OrderStatusCodes.Confirmed">@OrderStatusCodes.Confirmed</option>
+                    <option value="@OrderStatusCodes.Received">@OrderStatusCodes.Received</option>
+                    <option value="@OrderStatusCodes.Finalized">@OrderStatusCodes.Finalized</option>
+                    <option value="@OrderStatusCodes.Accepted">@OrderStatusCodes.Accepted</option>
+                    <option value="@OrderStatusCodes.Complete">@OrderStatusCodes.Complete</option>
```
srkirkland: @laholstege I think here either create a select list in order status codes or create a new array property (call it All?) that is just an array of all valid status codes.  Could then easily extend to have other arrays later.  That way on both of these pages you can keep your select list but then foreach your <option> tags so that we don't have to know where to change things when new statuses are added

Anlab.Mvc/Views/TestItems/Index.cshtml
```diff
@@ -22,10 +22,10 @@
...
                 @Html.DisplayNameFor(model => model.Category) <select name="category" id="category">
                     <option value="All">All</option>
-                    <option value="Soil">Soil</option>
-                    <option value="Plant">Plant</option>
-                    <option value="Water">Water</option>
-                    <option value="Other">Other</option>
+                    <option value="@TestCategories.Soil">@TestCategories.Soil</option>
+                    <option value="@TestCategories.Plant">@TestCategories.Plant</option>
+                    <option value="@TestCategories.Water">@TestCategories.Water</option>
+                    <option value="@TestCategories.Other">@TestCategories.Other</option>
```
srkirkland: @laholstege another good place where a preformatted array or select list would save duplication and future proof this better.

PR #182, opened 2017-09-11
Lh/pacific time

laholstege:


Anlab.Mvc/Views/Admin/MailQueue.cshtml
```diff
@@ -53,13 +53,13 @@
                 </div>
             </td>
             <td>
-                Created At: @message.CreatedAt.ToLocalTime()<br />
+                Created At: @DateTimeExtensions.ToPacificTime(message.CreatedAt).ToString("g")<br />
```
srkirkland: @laholstege since they are extensions, you don't need to call directly.  Fix this up to `message.CreatedAt.ToPacificTime().ToString("g")`

PR #185, opened 2017-09-12
Jcs/sloth check

jSylvestre:


Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+            }
+
+
+            builder.AddEnvironmentVariables();
+            Configuration = builder.Build();
+            
+
+            IServiceCollection services = new ServiceCollection();
+            services.AddDbContext<ApplicationDbContext>(options =>
+                    options.UseSqlite("Data Source=..\\..\\..\\..\\Anlab.Mvc\\bin\\Debug\\netcoreapp1.1\\anlab.db")
```
srkirkland: @jSylvestre Guess this is ok for now but it'll break when we hit 2.0 soon.  Maybe it would be better to make a /data folder or the old App_Data could also work, and have it live there

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+        public static async Task<bool> ProcessOrders()
+        {
+            var dbContext = Provider.GetService<ApplicationDbContext>();
+            var token = Configuration.GetSection("MoneyMovement:SlothApiKey").Value;
+            var url = Configuration.GetSection("MoneyMovement:SlothApiUrl").Value;
+
+
+            using (var client = new HttpClient())
+            {
+                var orders = dbContext.Orders.Include(i => i.ApprovedPayment).Where(a =>
```
srkirkland: @jSylvestre move this part outside of the httpclient.  No need to create the client if orders.Count == 0

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                    a.ApprovedPayment != null && a.Paid && a.Status != OrderStatusCodes.Complete).ToList();
+                client.BaseAddress = new Uri(url);
+                client.DefaultRequestHeaders.Add("X-Auth-Token", token);
+
+                if (orders?.Count > 0)
```
srkirkland: @jSylvestre just curious, but this can never be null right?  A .ToList() will never return null AFAIK so why do you need the `?`

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                            continue;
+                        }
+                        if (response.StatusCode == HttpStatusCode.NoContent)
+                        {
+                            continue;
+                        }
+                        if (response.IsSuccessStatusCode)
+                        { 
+                            var slothResponse = await response.GetContentOrNullAsync<Sloth>();
+                            slothResponse.JsonDetails = await response.Content.ReadAsStringAsync(); //TODO: Decide if we really need this.
```
srkirkland: @jSylvestre Yea I don't think we need to care about any of this, maybe we'll store the KFS Key but we don't particularly need that either since we can just look it up based on transaction_id.

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                    Console.WriteLine($"Updated {updatedCount} orders");
+                }
+
+            }
+
+            return true;
+
+
+
+            //try
```
srkirkland: @jSylvestre ✂️

Anlab.Jobs.MoneyMovement/appsettings.json
```diff
@@ -0,0 +1,5 @@
+﻿{
+  "MoneyMovement": {
```
srkirkland: @jSylvestre I would generalize as "Financial" so we can use along with other financial vars and secure them all the same way

PR #186, opened 2017-09-12
Migrate to .net core 2

srkirkland:


Anlab.Core/Data/ApplicationDbContext.cs
```diff
@@ -79,6 +79,14 @@ protected override void OnModelCreating(ModelBuilder builder)
             // Customize the ASP.NET Identity model and override the defaults if needed.
             // For example, you can rename the ASP.NET Identity table names and more.
             // Add your customizations after calling base.OnModelCreating(builder);
+
+            // prevent duplicate FKs when running migrations
```
srkirkland: @jpknoll couldn't figure out how to get this to work.  https://docs.microsoft.com/en-us/aspnet/core/migration/1x-to-2x/identity-2x suggests you can do this and then your user.roles relationship will still work, but it didn't for me, so I'm using the userManager service instead

Anlab.Mvc/AnlabMvc.csproj
```diff
@@ -14,32 +14,33 @@
...
-    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer.Design" Version="1.1.1" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="1.1.1" />
-    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="1.1.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="1.1.1" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="1.1.1" />
-    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="1.1.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
+    <PackageReference Include="Markdig" Version="0.13.2" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Authentication.Cookies" Version="2.0.0" />
```
srkirkland: @jpknoll most of this can go away into`Microsoft.AspNetCore.All` but I would have to go through each one to find out for sure.

Anlab.Mvc/Startup.cs
```diff
@@ -66,17 +71,69 @@ public void ConfigureServices(IServiceCollection services)
...
+                        if (identity == null)
+                        {
+                            return;
+                        }
+
+                        // email comes across in both name claim and upn
+                        var email = identity.FindFirst(ClaimTypes.Upn).Value;
+
+                        // look up user info and add as claims
+                        var user = await _directorySearchService.GetByEmail(email);
```
srkirkland: @jpknoll would like to get this service here instead of going the field route, but not sure how I can do it through the DI service

Anlab.Mvc/Controllers/AccountController.cs
```diff
@@ -90,8 +86,123 @@ public class AccountController : Controller
...
+        [HttpGet]
+        [AllowAnonymous]
+        public async Task<IActionResult> LoginWith2fa(bool rememberMe, string returnUrl = null)
+        {
+            // Ensure the user has gone through the username & password screen first
+            var user = await _signInManager.GetTwoFactorAuthenticationUserAsync();
+
+            if (user == null)
+            {
+                throw new ApplicationException($"Unable to load two-factor authentication user.");
```
jpknoll: Unnecessary $

Anlab.Mvc/Controllers/AdminController.cs
```diff
@@ -35,23 +35,24 @@ public AdminController(ApplicationDbContext dbContext, UserManager<User> userMan
         [Authorize(Roles = RoleCodes.Admin)]
         public async Task<IActionResult> Index()
         {
-            var adminRole = await _roleManager.FindByNameAsync(RoleCodes.Admin);
-            var userRole = await _roleManager.FindByNameAsync(RoleCodes.User);
+            // TODO: find better way than super select
+            var usersInRoles = _dbContext.Users.Select(u => new UserRolesModel { User = u });
```
jpknoll: I'm certain you could extend the IdentityRole model to include a Users property.

jpknoll: Or maybe they have a RoleManager?

Anlab.Mvc/Startup.cs
jpknoll: Yeah, don't do this here. Do it on the AccountController.ExternalLoginCallback.

Anlab.Mvc/Startup.cs
```diff
@@ -116,49 +174,49 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
-                }
-            };
-            app.UseCasAuthentication(casOptions);
-
-            app.UseGoogleAuthentication(new GoogleOptions()
-            {
-                ClientId = Configuration["Authentication:Google:ClientId"],
-                ClientSecret = Configuration["Authentication:Google:ClientSecret"]
-            });
+            //var casOptions = new CasOptions
```
jpknoll: ... Delete this

Anlab.Mvc/Controllers/AccountController.cs
srkirkland: @jpknoll yea, don't know why the template does that https://github.com/aspnet/Templates/blob/dev/template_feed/Microsoft.DotNet.Web.ProjectTemplates.2.0/content/StarterWeb-CSharp/Controllers/AccountController.cs#L150

Anlab.Mvc/Controllers/AdminController.cs
srkirkland: @jpknoll Yea I tried w/ the role manager and that worked as well, but this was slightly more efficient since otherwise we'd have to enumerate each role (instead of enumerating each user).  But maybe that's a better way to go since most of what we care about is just admin role users

Anlab.Mvc/Startup.cs
srkirkland: @jpknoll yea this i can delete now that the external login callback is here

Anlab.Mvc/Extensions/EmailSenderExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+using AnlabMvc.Services;
+
+namespace AnlabMvc.Extensions
+{
+    public static class EmailSenderExtensions
+    {
+        public static Task SendEmailConfirmationAsync(this IEmailSender emailSender, string email, string link)
+        {
+            return emailSender.SendEmailAsync(email, "Confirm your email",
+                $"Please confirm your account by clicking this link: <a href='{HtmlEncoder.Default.Encode(link)}'>link</a>");
```
jSylvestre: @srkirkland We are not using this because we only allow google and ucdavis. Correct? This would be is they were going to create a username/password?

srkirkland: @jpknoll correct -- We can re-enable in the case we ever decide to handle username/pass, but for now there will be no way for them to use this since we only support those 2 you mentioned

PR #188, opened 2017-09-12
Jcs/pay20170906

jSylvestre:
I think this will need re-working. Just creating the PR so I don't forget about it


jSylvestre:
Going to do this a different way...

PR #189, opened 2017-09-12
add confirmed page

laholstege:


Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -235,9 +235,32 @@ public OrderController(ApplicationDbContext context, IOrderService orderService,
...
+            if (order.CreatorId != CurrentUserId)
+            {
+                ErrorMessage = "You don't have access to this order.";
+                return NotFound();
+            }
+
+            var model = new OrderReviewModel();
+            model.Order = order;
+            model.OrderDetails = order.GetOrderDetails();
+            model.TestItems = _context.TestItems
```
srkirkland: @laholstege do we really need to put the testItems here?  aren't they stored in the order details already?  If we don't need it we can save an expensive deserialization step

PR #193, opened 2017-09-14
Do i base design or base master

cydoval:


srkirkland:
closing because we got the PR the wrong way

PR #195, opened 2017-09-14
fix adjustment amount

### LINKED ISSUE #191: Adjustment Amount
Need to default the Adjust Final Total: amount when the finalize page is opened.
And/Or, need a validation that a blank value isn't allowed.

Otherwise, if they go in and finalize without entering anything it zeros out the grand total.

laholstege:
fixes #191


Anlab.Mvc/Views/Lab/UpdateFromCompletedTests.cshtml
```diff
@@ -56,6 +57,8 @@
                 var adjustAmt = parseFloat(grandTotal) - parseFloat(prevTotal);
             else
                 var adjustAmt = 0;
+            console.log(adjustAmt);
```
srkirkland: @laholstege dump this

PR #196, opened 2017-09-14
Lh/more info from fake labworks db

laholstege:


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -119,9 +119,23 @@ public IActionResult ListUsersOrders(string id)
                 ErrorMessage = "A request number is required";
                 return RedirectToAction("Confirmation");
             }
+            order.RequestNum = requestNum;
+            var result = await _orderService.OverwiteOrderFromDb(order); //TODO: Just testing
```
srkirkland: @laholstege this will probably have to change since there could be other errors, like work request number not being found.  But for now it's fine, let's just remember to come back here after the SQL DB is back up

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -143,13 +157,16 @@ public IActionResult ListUsersOrders(string id)
                 return NotFound();
             }
 
-            var result = await _orderService.OverwiteOrderWithTestsCompleted(order); //TODO: Just testing
+            var result = await _orderService.OverwiteOrderFromDb(order); //TODO: Just testing
```
srkirkland: @laholstege what's all the TODO's about?

laholstege: whoops I just didn't push the commit that changed them

PR #205, opened 2017-09-18
Lh/react form errors

laholstege:


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -203,18 +203,20 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
 
                     <SampleTypeSelection sampleType={sampleType} onSampleSelected={this.onSampleSelected} />
 
-                    <div className="form_wrap">
+                    <div className="form_wrap" id="quantity">
```
srkirkland: @laholstege move the id to the label below instead of this div

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -203,18 +203,20 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
...
                     <SampleTypeSelection sampleType={sampleType} onSampleSelected={this.onSampleSelected} />
 
-                    <div className="form_wrap">
+                    <div className="form_wrap" id="quantity">
                         <label className="form_header">How many samples will you require?</label>
                         <Quantity quantity={quantity} onQuantityChanged={this.onQuantityChanged} />
                     </div>
                     <AdditionalEmails addedEmails={additionalEmails} onEmailAdded={this.onEmailAdded} onDeleteEmail={this.onDeleteEmail}/>
-                    <Project project={project} handleChange={this.handleChange} />
+                    <div id="project">
```
srkirkland: @laholstege this one is ok unless you feel like adding a label describing what project is (which will be good eventually)

Anlab.Mvc/Client/components/Summary.tsx
```diff
@@ -56,6 +56,18 @@ export class Summary extends React.Component<ISummaryProps, any> {
...
+        let errorIconStyle = {
+            color: 'red',
+            padding: '0 10px'
+        }
+        let errorLink = null;
+        if (!this.props.hideError)
+        {
+            if (this.props.quantity < 1)
+                errorLink = <a href="#quantity">Fix Errors<i className="fa fa-exclamation" style={errorIconStyle} ></i></a>
+            else
```
srkirkland: @laholstege let's not assume these will be the only two errors, so throw an IF into this else and do a `else if (!this.props.project)` so we know project is empty before showing that as the error

PR #206, opened 2017-09-19
Jcs/money movement20170915

jSylvestre:


jSylvestre:
Still need to call it from an actual method in the payment controller.

Anlab.Core/Services/SlothService.cs
```diff
@@ -0,0 +1,131 @@
...
+        //TODO: Add validation?
+        public async Task<SlothResponseModel> MoveMoney(Order order)
+        {
+            var orderDetails = order.GetOrderDetails();
+            var token = _appSettings.SlothApiKey;
+            var url = _appSettings.SlothApiUrl;
+            var creditAccount = new AccountModel(_appSettings.AnlabAccount);
+            var debitAccount = new AccountModel(orderDetails.Payment.Account);
+
+            var objectCode = _appSettings.ObjectCode;
```
jSylvestre: @srkirkland Would the object code be used on both the Debit and Credit?

jSylvestre: nm, I see it does

Anlab.Core/Domain/Order.cs
```diff
@@ -52,9 +52,13 @@ public Order()
         public PaymentEvent ApprovedPayment { get; set; }
         
         public string KfsTrackingNumber { get; set; }
+        public string SlothTransactionId { get; set; } //TODO: Change to Guid? ?
 
         public bool Paid { get; set; } = false;
 
+        public bool IsInternalClient { get; set; } = false;
```
srkirkland: @jSylvestre Ah i was thinking to replace this with a single field like string PaymentType which would be like status.  values of "CreditCard", "UCDAccount" "IOCAccount" and "Other" perhaps.  Plus it could cover other cases for us later too.

Anlab.Core/Models/AccountModel.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Anlab.Jobs.MoneyMovement
+{
+    public class AccountModel
+    {
+        public AccountModel(string rawAccount)
+        {
+            var temp = rawAccount.Split('-');
```
srkirkland: @jSylvestre might be worth throwing in a regex or at least checking that temp has length 2 before doing the temp[1] index, otherwise this will throw an exception on a bad account instead of just returning null account info.

srkirkland: @jSylvestre I know regex is no fun but with capture groups this whole thing could be one line w/ the regex, which would also do your length validation at the same time.   something like `(\w)-(\w{7})\/?(\w{5})?` which is basically 3 capture groups, one for chart, one for 7 length account, then an optional slash and an optional group for subacct.  Basically just plug in and use Regex.Matches to take care of it

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -41,70 +39,30 @@ static void Main(string[] args)
...
             );
+            services.AddTransient<ISlothService, SlothService>();           
             Provider = services.BuildServiceProvider();
+            
+            SlothService = Provider.GetService<ISlothService>();           
 
-            Console.WriteLine("About to start");
-            var result = Task.Run(() => ProcessOrders()).Result; //Wasn't able to debug this unless it returned a result...
+            Console.WriteLine("About to start Credit Cards");
+            var financialSettings = new FinancialSettings
```
srkirkland: @jSylvestre why did you get away from a separate method that's called from Main() in favor of doing Task.Run directly in main?

Anlab.Core/Domain/Order.cs
srkirkland: @jSylvestre Looking through the rest of the code I guess this would be a big change, so maybe don't worry about this part

Anlab.Jobs.MoneyMovement/Program.cs
jSylvestre: @srkirkland I just liked it all grouped in the sloth service. If I wasn't going to call the service from the MVC side too, I would have left it here.

PR #209, opened 2017-09-19
Lh/add emails

### LINKED ISSUE #190: client email entry for anlab's "copy-to"
Once we decide we want to allow multiple, let's go with something like http://prakhar1989.github.io/react-tags/example/.

laholstege:
closes #190

### DIFF
--- a/Anlab.Mvc/Client/components/AdditionalEmails.tsx
+++ b/Anlab.Mvc/Client/components/AdditionalEmails.tsx
@@ -1,6 +1,7 @@
 ﻿import * as React from 'react';
 import Input from 'react-toolbox/lib/input';
 import { Button, IconButton } from 'react-toolbox/lib/button';
+import Chip from 'react-toolbox/lib/chip';
 
 interface IAdditionalEmailsProps {
     addedEmails: Array<string>;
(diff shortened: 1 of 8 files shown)


Anlab.Mvc/Client/components/AdditionalEmails.tsx
```diff
@@ -1,15 +1,20 @@
...
     addedEmails: Array<string>;
     onEmailAdded: Function;
     onDeleteEmail: Function;
+    defaultEmail: string;
 }
 
 interface IAdditionalEmailsState {
     email: string;
+    toggle: boolean;
+    error: boolean;
```
srkirkland: @laholstege since it's bool update the name to something like `HasError` or anything that has binary meaning, since otherwise error and errorText are a bit ambiguous

Anlab.Mvc/Client/components/AdditionalEmails.tsx
```diff
@@ -52,30 +61,59 @@ export class AdditionalEmails extends React.Component<IAdditionalEmailsProps, IA
...
+            if (this.state.email === "")
+                this._toggleAddEmail();
+            else
+                this.onClick();
+        }
+    }
+
+    handleBlur = () => {
+        if (this.state.email != "") {
+            this.onClick();
```
srkirkland: @laholstege nit: extract the onClick method to a new method, maybe `_addEmailAddress` or similar, then call from other methods.  That way you don't have to call this.onClick() even for things like keypress and blur that isn't really clicks

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -207,7 +207,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
                         <label className="form_header" id="quantity">How many samples will you require?</label>
                         <Quantity quantity={quantity} onQuantityChanged={this.onQuantityChanged} />
                     </div>
-                    <AdditionalEmails addedEmails={additionalEmails} onEmailAdded={this.onEmailAdded} onDeleteEmail={this.onDeleteEmail}/>
+                    <AdditionalEmails addedEmails={additionalEmails} onEmailAdded={this.onEmailAdded} onDeleteEmail={this.onDeleteEmail} defaultEmail={window.App.defaultEmail} />
```
srkirkland: @laholstege can you set the defaultEmail to this.state for the order form component instead of using window here, so we can pin window.App access to just the ctor.

PR #213, opened 2017-09-20
Jcs/account regex

jSylvestre:
image


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -48,7 +48,7 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
         const initialState = {
             orderId: null,
             additionalInfo: '',
-            payment: { clientType: 'uc', account: '' },
+            payment: { clientType: 'C', account: '' },
```
srkirkland: @jSylvestre why are we not defaulting to UC client?

jSylvestre: @srkirkland I don't remember that. I'll look to see when/why it was changed.

jSylvestre: @srkirkland My bad. Changed by mistake. But it gets overwritten below. We default to uc if they have an account in their profile. (Added the regex to the user account too)

PR #217, opened 2017-09-21
Check if the UCD Account money has moved to the final destination.

jSylvestre:
Need to get what the status value is

### DIFF
--- a/Anlab.Core/Services/SlothService.cs
+++ b/Anlab.Core/Services/SlothService.cs
@@ -113,11 +113,17 @@ namespace Anlab.Core.Services
                     {
                         var content = await response.Content.ReadAsStringAsync();
                         var slothResponse = JsonConvert.DeserializeObject<SlothResponseModel>(content);
-                        if (slothResponse.Status == "???")
+                        if (slothResponse.Status == "Completed")
                         {
                             updatedCount++;
                             order.Status = OrderStatusCodes.Complete;
-                        }                     
+                        }
+                        if (slothResponse.Status == "Cancelled")
+                        {
+                            order.Paid = false;
+                            Console.WriteLine($"Order {order.Id} was cancelled. Setting back to unpaid");
+                            //TODO: Write to the notes field? Trigger off an email?
+                        }
                         
                     }
                 }
--- a/Anlab.Core/Services/SlothService.cs
+++ b/Anlab.Core/Services/SlothService.cs
@@ -16,9 +16,9 @@ namespace Anlab.Core.Services
     public interface ISlothService
     {
         Task<SlothResponseModel> MoveMoney(Order order);
-        Task<bool> ProcessCreditCards(FinancialSettings financialSettings);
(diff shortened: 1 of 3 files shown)


Anlab.Core/Services/SlothService.cs
```diff
@@ -17,6 +17,8 @@ public interface ISlothService
     {
         Task<SlothResponseModel> MoveMoney(Order order);
         Task<bool> ProcessCreditCards(FinancialSettings financialSettings);
+
+        Task<bool> MoneyMovedCheck(FinancialSettings financialSettings);
```
srkirkland: @jSylvestre These methods that return a bool would be good to have boolean sounding names, maybe like `MoneyHasMoved` or `VerifyMoneyHasMoved`, etc, for example.  Maybe for the processing method you don't need it, but this money check seems like a obvious one.

Anlab.Core/Services/SlothService.cs
```diff
@@ -77,6 +81,59 @@ public SlothService(ApplicationDbContext dbContext, IOptions<FinancialSettings>
             return new SlothResponseModel { Success = false };
         }
 
+        public async Task<bool> MoneyMovedCheck(FinancialSettings financialSettings)
+        {
+            Console.WriteLine("Beginning UCD money movement check");
```
srkirkland: @jSylvestre remind me to help add real logging soon

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -61,6 +61,12 @@ static void Main(string[] args)
             }
             Console.WriteLine("Done Credit Cards");
 
+            result = Task.Run(() => SlothService.MoneyMovedCheck(financialSettings)).Result;
```
srkirkland: @jSylvestre sorta just FYI but if you want you can run these two in parallel since they are orthogonal, using Task.WhenAll().

Anlab.Jobs.MoneyMovement/Program.cs
```diff
@@ -61,6 +61,12 @@ static void Main(string[] args)
             }
             Console.WriteLine("Done Credit Cards");
 
+            result = Task.Run(() => SlothService.MoneyMovedCheck(financialSettings)).Result;
+            if (!result)
```
srkirkland: @jSylvestre what's the point of returning the bool, wouldn't that be something you could just as easily log within the method itself?

jSylvestre: @srkirkland When I was debugging this it wouldn't debug into the method without it.

jSylvestre: @srkirkland I can do this, then remove the return value:
            SlothService.ProcessCreditCards(financialSettings).GetAwaiter().GetResult();
            SlothService.MoneyHasMoved(financialSettings).GetAwaiter().GetResult();

PR #218, opened 2017-09-21
LH/change edit to also save prices

laholstege:


jSylvestre:
@laholstege Was this supposed to update when the confirm it too?

Anlab.Mvc/Controllers/OrderController.cs
```diff
@@ -231,8 +241,19 @@ public OrderController(ApplicationDbContext context, IOrderService orderService,
                 return RedirectToAction("Index");
             }
 
-            order.Status = OrderStatusCodes.Confirmed;
+            var allTests = await _orderService.PopulateTestItemModel(true);
```
srkirkland: @laholstege this code is copied several times, and at the end it just essentially takes an order and returns an updated order -- maybe a combined method in the order service would save us from future copy errors?  `_orderService.UpdateTestsAndRecalculate(order)` or similar?

PR #221, opened 2017-09-21
Lh/add sop

laholstege:


srkirkland:
@laholstege Do we actually want/need to store the SOP field in the database?  Seems like it's the kind of thing that we'll pull from labworks and not want to save.

Anlab.Mvc/Client/components/TestList.tsx
```diff
@@ -75,8 +76,8 @@ export class TestList extends React.Component<ITestListProps, ITestListState> {
                             <Checkbox checked={selected} onChange={e => this.onSelection(item, e)} />
                         </td>
                         <td>{item.analysis}</td>
-                        <td>{item.id}</td>
-                        <td><NumberFormat value={priceDisplay} displayType={'text'} thousandSeparator={true} decimalPrecision={2} prefix={'$'} /></td>
+                        <td>{item.analysisMethodNumber}</td>
```
jSylvestre: @srkirkland Was this going to be a link or tooltip?

PR #226, opened 2017-09-25
allow xplat time zones

srkirkland:


srkirkland:
@jSylvestre Because not all systems have the "Pacific Standard Time" timezone loaded.  Namely mac and linux.

PR #235, opened 2017-09-25
Lh/kfs account lookup

### LINKED ISSUE #160: KFS Account Lookup
An attempt to have them verify the account they enter is the correct one

laholstege:
closes #160

### DIFF
--- a/Anlab.Mvc/Client/components/PaymentSelection.tsx
+++ b/Anlab.Mvc/Client/components/PaymentSelection.tsx
@@ -45,7 +45,7 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
     lookupAccount = () => {
         if (!this.state.error && this.props.payment.account !== null && this.checkChart(this.props.payment.account.charAt(0)))
         {
-            fetch(`/financial/${this.props.payment.account}`, { credentials: 'same-origin' })
+            fetch(`/financial/info?account=${this.props.payment.account}`, { credentials: 'same-origin' })
                 .then(response => {
                     if (!response.ok) {
                         throw Error();
--- a/Anlab.Mvc/Client/components/PaymentSelection.tsx
+++ b/Anlab.Mvc/Client/components/PaymentSelection.tsx
@@ -38,10 +38,13 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
     lookupAccount = (account: string) => {
         if (this.state.error === "")
         {
-            console.log("fetching...");
             fetch(`/financial/${this.props.payment.account}`, { credentials: 'same-origin' })
                 .then(response => response.json())
                 .then(accountName => this.setState({ accountName }));
+        }
+        else
+        {
+            this.setState({ accountName: null });
         }
     }
 
--- a/Anlab.Mvc/Client/components/PaymentSelection.tsx
+++ b/Anlab.Mvc/Client/components/PaymentSelection.tsx
@@ -35,12 +35,22 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
         }
     }
 
-    lookupAccount = (account: string) => {
-        if (this.state.error === "")
+    lookupAccount = () => {
+        if (!this.state.error && this.props.payment.account !== null &&
+            (this.props.payment.account.charAt(0) === "L" || this.props.payment.account.charAt(0) === "l" || this.props.payment.account.charAt(0) === "3"))
         {
             fetch(`/financial/${this.props.payment.account}`, { credentials: 'same-origin' })
+                .then(response => {
+                    if (!response.ok) {
+                        throw Error();
+                    }
+                    return response;
+                })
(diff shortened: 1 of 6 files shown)


Anlab.Mvc/Client/components/PaymentSelection.tsx
```diff
@@ -17,18 +18,36 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
...
+                <div>
+                    <Input type="text" label="UC Account" error={this.state.error} value={this.props.payment.account} maxLength={15} onChange={this.handleAccountChange} onBlur={this.lookupAccount} />
+                    {this.state.accountName}
+                </div>
             );
         }
     }
+
+    lookupAccount = (account: string) => {
+        if (this.state.error === "")
```
srkirkland: @laholstege it's all the same but since empty string is falsey you can just do `if (!this.state.error)` and it'll handle other cases like null and undefined

Anlab.Mvc/Controllers/FinancialServiceController.cs
```diff
@@ -0,0 +1,30 @@
...
+        public FinancialServiceController(IFinancialService financialService)
+        {
+            _financialService = financialService;
+        }
+
+        [HttpGet("financial/{account}")]
+        public async Task<string> GetAccountInfo(string account)
+        {
+            var accountModel = new AccountModel(account);
+            var result = await _financialService.GetAccountInfo(accountModel.Chart, accountModel.Account);
```
srkirkland: @laholstege throw in a little null checking after you create the account model, since there is no reason to call the financial service w/ a null or empty chart or account.  In that case returning null seems easiest, though empty string works too.  Also call it GetAccountName() for now since it's only the name that we are returning.

Anlab.Mvc/Services/IFinancialService.cs
```diff
@@ -0,0 +1,28 @@
...
+        Task<string> GetAccountInfo(string chart, string account);
+    }
+
+    public class FinancialService: IFinancialService
+    {
+        public async Task<string> GetAccountInfo(string chart, string account)
+        {
+            using (var client = new HttpClient())
+            {
+                var url = "https://kfs.ucdavis.edu/kfs-prd/remoting/rest/fau/account/" + chart + "/" + account + "/name";
```
srkirkland: @laholstege use string format or interpolation instead of concatination

srkirkland: @laholstege also, eventually, this should go in an app setting.

Anlab.Mvc/Controllers/FinancialServiceController.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class FinancialServiceController : ApplicationController
+    {
+        private readonly IFinancialService _financialService;
+        
+        public FinancialServiceController(IFinancialService financialService)
+        {
+            _financialService = financialService;
+        }
+
+        [HttpGet("financial/{account}")]
```
jSylvestre: I'm not a big fan of having different routes from the controller/action.

Anlab.Mvc/Client/components/PaymentSelection.tsx
```diff
@@ -17,18 +18,36 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
...
-                <Input type="text" label="UC Account" error={this.state.error} value={this.props.payment.account} maxLength={15} onChange={this.handleAccountChange}/>
+                <div>
+                    <Input type="text" label="UC Account" error={this.state.error} value={this.props.payment.account} maxLength={15} onChange={this.handleAccountChange} onBlur={this.lookupAccount} />
+                    {this.state.accountName}
+                </div>
             );
         }
     }
+
+    lookupAccount = (account: string) => {
```
jSylvestre: @laholstege account param is not being used.

Anlab.Mvc/Controllers/FinancialServiceController.cs
jSylvestre: @laholstege Only call if it is a uc davis account (Chart 3 or L). @srkirkland Or does this get info for other colleges?

Anlab.Mvc/Client/components/PaymentSelection.tsx
```diff
@@ -17,18 +18,36 @@ export class PaymentSelection extends React.Component<IPaymentProps, any> {
...
+        if (this.state.error === "")
+        {
+            fetch(`/financial/${this.props.payment.account}`, { credentials: 'same-origin' })
+                .then(response => response.json())
+                .then(accountName => this.setState({ accountName }));
```
jSylvestre: @laholstege If it doesn't find it, clear out the account name:
So if I enter 3-BG13DOE, tab off (finds it), then change it to x-BG13DOE it doesn't find it and keeps the previous name.

PR #254, opened 2017-09-28
Lh/lab receive adjustment

### LINKED ISSUE #231: /Lab/Confirmation needs comments and adjustment amount/total
In the email we sent someone after their order has been received, we need to send along the comments and adjustment info so they know why the amount changed.

laholstege:
fixes #231

### DIFF
--- a/Anlab.Mvc/Controllers/LabController.cs
+++ b/Anlab.Mvc/Controllers/LabController.cs
@@ -103,15 +103,15 @@ namespace AnlabMvc.Controllers
         }
 
         [HttpPost]
-        public async Task<IActionResult> Confirmation(int id, bool confirm, string requestNum)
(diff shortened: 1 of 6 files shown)


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -148,6 +149,60 @@ public IActionResult ListUsersOrders(string id)
...
+
+            if (order == null)
+            {
+                return NotFound();
+            }
+
+            var model = new OrderReviewModel();
+            model.Order = order;
+            model.OrderDetails = order.GetOrderDetails();
+            model.HideLabDetails = false;
```
srkirkland: @laholstege not being used anywhere anymore right?

PR #260, opened 2017-10-03
Don't do async in ctor

srkirkland:


srkirkland:
Might not need this but we shouldn't be doing async in the ctor anyway

PR #264, opened 2017-10-04
Link to the SOP page from the order

jSylvestre:


Anlab.Mvc/Client/components/TestList.tsx
```diff
@@ -70,13 +70,15 @@ export class TestList extends React.Component<ITestListProps, ITestListState> {
                 const selected = !!this.props.selectedTests[item.id];
                 const priceDisplay = (this.props.payment.clientType === 'uc' ? item.internalCost : item.externalCost);
                 const tooltipContent = converter.makeHtml(item.notes);
+                const url = '/analysis/' + item.category + '/' + item.sop;
```
srkirkland: @jSylvestre more out of curiosity than anything else, does string interpolation work here in typescript?  So it would be 

            const url = `/analysis/${item.category}/${item.sop}`;

jSylvestre: @srkirkland Interesting. works with ` but not '

srkirkland: @jSylvestre correct, backticks are required for string interpolation, otherwise they are just regular strings.  JavaScript history i supposed

PR #265, opened 2017-10-04
Design 9

cydoval:
quite a few layout changes mainly to receipt and order details


jSylvestre:
@cydoval @srkirkland I think I got everything fixed up. It might be worthwhile to actually go through the pages to make sure they look like what you want. I also posted a picture in #anlab in slack.

PR #268, opened 2017-10-05
Put the form for Finalize above the table.

jSylvestre:
image


jSylvestre:
Cal will fix on his branch.

PR #269, opened 2017-10-05
Lh/special test modal

laholstege:


jSylvestre:
@srkirkland @laholstege This still needs to be displayed and printed?

srkirkland:
@jSylvestre yea this is nowhere near done/ready

srkirkland:
@laholstege ok I think I fixed everything up so the model will bind correctly in both directions, and then save the result into additional info property.  We still have the issue on the model opening when doing edit, but i think you had a quick fix for that.

PR #275, opened 2017-10-09
Redirect to profile edit page

jSylvestre:
Tested with UC Davis Login, but not google


jSylvestre:
@srkirkland Adding the question here because it is for this cs file.
Should we try and grab the client id and default account when we create the user?

PR #288, opened 2017-10-12
Lh/client id info modal

laholstege:


Anlab.Mvc/Client/components/ClientIdModal.tsx
```diff
@@ -0,0 +1,100 @@
...
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            newClientInfo: {
+                employer: this.props.clientInfo.employer,
+                name: this.props.clientInfo.name,
+                email: this.props.clientInfo.email,
+                phoneNumber: this.props.clientInfo.phoneNumber
+            },
```
srkirkland: @laholstege since state.newClientInfo is the same type as props.clientInfo, I think you can make this whole thing into `newClientInfo: ...this.props.clientInfo` which just spreads and copies the properties.

Also just to keep in mind, though it doesn't hurt anything here -- usually copying props into state is not great since if props.clientInfo ever changed, this component would not update to reflect the changes.  Again it's ok here since we know props.clientInfo doesn't change outside of this component, but just something to remember

Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -131,6 +140,17 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
         this.setState({ ...this.state, quantity }, this.validate);
     }
 
+    updateNewClientInfo = (info: INewClientInfo) => {
+        this.setState({...this.state,
+            newClientInfo: {
```
srkirkland: @laholstege if you wanted to here's another change to make this a one-liner using the spread operator since the info param passed in is of the same type as your state.

Anlab.Mvc/Views/AdminAnalysis/_AnalysisMethodForm.cshtml
```diff
@@ -21,4 +46,5 @@
         <textarea asp-for="Content" class="form-control" rows="3"> </textarea>
         <span asp-validation-for="Content" class="text-danger"></span>
     </div>
+>>>>>>> master
```
srkirkland: @laholstege Looks like you have some unresolved merge conflict in here

PR #293, opened 2017-10-16
Data Annotation Updates

jSylvestre:
Put tests back


Anlab.Core/Services/SlothService.cs
```diff
@@ -100,7 +100,7 @@ public SlothService(ApplicationDbContext dbContext, IOptions<FinancialSettings>
                 var roledBackCount = 0;
                 foreach (var order in orders)
                 {
-                    var response = await client.GetAsync(order.SlothTransactionId);
+                    var response = await client.GetAsync(order.SlothTransactionId.ToString());
```
srkirkland: @jSylvestre since you changed it to a nullable Guid, if you are going to use it you should check to make sure it HasValue before calling the API here.

Issue #286, opened 2017-10-12
Soil: additional questions


srkirkland:
1. Is your sample imported?  yes/no (default to no).

Issue #285, opened 2017-10-12
Plant: additional questions


srkirkland:
Need to ask this question, which has 3 options:

Reporting basis:
1. [DEFAULT] Report results on 100% dry weight basis, based on an average of 10% of the samples.
1. Report on As Received bases
1. Report results on 100% dry weight basis, based on individual dry matter results (charge applies)

PR #305, opened 2017-10-19
Lh/sample specific questions

### LINKED ISSUE #284: Water: additional questions
Need to ask these questions

1. Are these samples filtered?  yes/no (default no)
1. Preservative Added? yes/no (default no).
        * if yes, the must specify w/ textbox
1. Cl and Soluble Ca, Mg, and Na are reported in meq/L.  Select here if you want them reported in mg/L.  (default no)

laholstege:
closes #284 #285 #286


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -118,7 +137,8 @@ export default class OrderForm extends React.Component<undefined, IOrderState> {
         this.state = { ...initialState };
     }
     validate = () => {
-        let valid = this.state.quantity > 0 && this.state.quantity <= 100 && !!this.state.project.trim();
+        let valid = this.state.quantity > 0 && this.state.quantity <= 100 && !!this.state.project.trim()
+            && (this.state.sampleType != "Water" || !this.state.sampleTypeQuestions.waterPreservativeAdded || !!this.state.sampleTypeQuestions.waterPreservativeInfo.trim());
```
srkirkland: @laholstege use tripple equals (!==)

Anlab.Mvc/Client/components/SampleSoilQuestions.tsx
```diff
@@ -0,0 +1,48 @@
...
+    sampleType: string;
+    questions: ISampleTypeQuestions;
+}
+
+export class SampleSoilQuestions extends React.Component<ISampleSoilQuestions, {}> {
+
+    constructor(props) {
+        super(props);
+
+    }
```
srkirkland: @laholstege you don't need any of these if they are just empty like this

jSylvestre:
@srkirkland @laholstege Do we care that we have non soil type values saved?

sampleTypeQuestions: {
soilImported: true,
plantReportingBasis: "Report results on 100% dry weight basis, based on an average of 10% of the samples.",
waterFiltered: false,
waterPreservativeAdded: false,
waterPreservativeInfo: null,
waterReportedInMgL: false
},

jSylvestre:
I'd say ignore this bug when you can save an order without the extra water text, but here are the steps if you want to reproduce it:
1) Create order, choose water, select yes for preservative but don't enter text
2) Choose soil and save
3) Edit order
4) Choose water and a water test
5) save

sampleTypeQuestions: {
soilImported: false,
plantReportingBasis: "Report results on 100% dry weight basis, based on an average of 10% of the samples.",
waterFiltered: false,
waterPreservativeAdded: true,
waterPreservativeInfo: null,
waterReportedInMgL: false
},

PR #333, opened 2017-10-30
Jcs/admin actions

### LINKED ISSUE #326: Way for admins to modify orders
Set order to any status, set paid flag, also remove/inactivate in case of duplicates (or make a removed status and set that)

jSylvestre:
Closes #326


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -325,6 +325,55 @@ public IActionResult ListUsersOrders(string id)
...
+            orderToUpdate.IsDeleted = order.IsDeleted;
+            if (order.IsDeleted)
+            {
+                ErrorMessage = "Order deleted!!!";
+            }
+
+            await _dbContext.SaveChangesAsync();
+
+            Message = "Order Updated";
+            if (order.IsDeleted)
```
srkirkland: @jSylvestre why do you have the same if statement twice.  is one of them supposed to return?  And why a message plus an ErrorMessage for the isDeleted step

jSylvestre: My bad...

PR #336, opened 2017-10-31
Lh/order form fixes

laholstege:
replace rest of Inputs
required either ClientId or NewClientInfo to be filled out
if UC account require account number
fix error button focuses on client id/uc account
formatting


jSylvestre:
@laholstege FYI, you need a client id to save, but it doesn't have to be valid.
You can test this by using fail as the client id.

PR #344, opened 2017-11-06
Public prices page

### LINKED ISSUE #327: Public page which shows all tests and prices

jSylvestre:
Closes #327

### DIFF
--- a/Anlab.Mvc/Views/Prices/Index.cshtml
+++ b/Anlab.Mvc/Views/Prices/Index.cshtml
@@ -9,22 +9,26 @@
     <table class="table">
         <thead>
         <tr>
-            <th>@Html.DisplayNameFor(model => model[0].Category)</th>
-            <th>@Html.DisplayNameFor(model => model[0].Group)</th>
-            
-            <th>@Html.DisplayNameFor(model => model[0].Analysis)</th>
+            <th rowspan="2">@Html.DisplayNameFor(model => model[0].Category)</th>
+            <th rowspan="2">@Html.DisplayNameFor(model => model[0].Group)</th>
+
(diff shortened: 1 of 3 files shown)


jSylvestre:
image

Anlab.Mvc/Views/Prices/Index.cshtml
```diff
@@ -0,0 +1,60 @@
...
+        <tr>
+            <th>@Html.DisplayNameFor(model => model[0].Category)</th>
+            <th>@Html.DisplayNameFor(model => model[0].Group)</th>
+            
+            <th>@Html.DisplayNameFor(model => model[0].Analysis)</th>
+            @*<th>@Html.DisplayNameFor(model => model[0].Id)</th>*@
+            <th>@Html.DisplayNameFor(model => model[0].InternalCost).Humanize()</th>
+            <th>@Html.DisplayNameFor(model => model[0].InternalSetupCost).Humanize()</th>
+            <th>@Html.DisplayNameFor(model => model[0].ExternalCost).Humanize()</th>
+            <th>@Html.DisplayNameFor(model => model[0].ExternalSetupCost).Humanize()</th>
```
srkirkland: @jSylvestre Change the header values for the last 4 columns to represent UC/non-UC rates.  Use a colspan to span a major grouping, like so

<thead><tr><th colspan="3"></th><th colspan="2">UC Clients</th><th colspan="2">Non-UC Clients</th></tr><tr>[...your original tr except drop internal/external]</tr></thead>

That should give it a nice tiered look and also group by what they understand, assuming datatables is ok with that.  If it isn't, just change the name to UC Cost instead of Internal Cost, etc, etc.

jSylvestre: @srkirkland Like this?

image

PR #355, opened 2017-11-09
add in logging

srkirkland:
Change the way stackify logging is setup to the .net core 2 way.  Plus I added a Jobs.Core project which is currently just a logging helper, then I use that in both jobs to setup logging


srkirkland:
closes #355

PR #358, opened 2017-11-15
WIP: Srk/logging

laholstege:
Merge logging branch


laholstege:
@jSylvestre I reverted to before your merge and did the conflict resolution myself.  You'll need to redo the changing of default connection string and whatever else you did in those 2 commits. This PR should work now and only change the files we care about

PR #375, opened 2017-11-17
Lh/order form browsing

laholstege:
add browse mode to view prices without having to add all the info required for an order


jSylvestre:
image

srkirkland:
We should change it so you can't click Save & Review if in browsing mode

PR #384, opened 2017-11-30
Le fancy checkbox

jpknoll:


jpknoll:
@laholstege There ya go...

jSylvestre:
@jpknoll When I run this, I can't save the order. I'm guessing it is because it is not detecting the tests being selected

jpknoll:
Oh good spot. I'll give it another look.

PR #396, opened 2017-12-12
Link to an order's emails

jSylvestre:
Added order id to email queue table and have a nullable parameter to pass the order id.


jSylvestre:
image

PR #405, opened 2018-01-10
See what was different

jSylvestre:


srkirkland:
@jSylvestre can we close/dump this?

PR #413, opened 2018-01-18
Validate kerb

jSylvestre:
Because they might use a shared dept account (CAS Validates these)

### DIFF
--- a/Anlab.Mvc/Models/Person.cs
+++ b/Anlab.Mvc/Models/Person.cs
@@ -8,4 +8,11 @@ namespace AnlabMvc.Models
         public string Mail { get; internal set; }
         public string FullName { get; internal set; }
     }
+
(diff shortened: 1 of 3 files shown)


jSylvestre:
@srkirkland simple version could hide other errors. I'll try the more complex one.

Anlab.Mvc/Models/Person.cs
```diff
@@ -8,4 +8,11 @@ public class Person
         public string Mail { get; internal set; }
         public string FullName { get; internal set; }
     }
+
+    public class ValidPerson
```
srkirkland: @jSylvestre why would a ValidPerson object have IsInvalid field?  Maybe change name to DirectoryPerson?  Or maybe just DirectoryResult, since that returns the result plus an associated person.

PR #423, opened 2018-01-29
Do not change if clicking on the active button.

### LINKED ISSUE #415: Clicking create order / browse should not toggle
it should just select whichever one you clicked

jSylvestre:
close #415

### DIFF
--- a/Anlab.Mvc/Client/components/ViewMode.tsx
+++ b/Anlab.Mvc/Client/components/ViewMode.tsx
@@ -16,14 +16,14 @@ export class ViewMode extends React.Component<IViewModeProps, any> {
                 <div className="row">
                     <div
                         className={this.props.placingOrder ? activeDiv : inactiveDiv}
-                        onClick={() => this._handleChange()}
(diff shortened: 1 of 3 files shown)


Anlab.Mvc/Client/components/ViewMode.tsx
```diff
@@ -32,8 +32,10 @@ export class ViewMode extends React.Component<IViewModeProps, any> {
         );
     }
 
-    private _handleChange = () => {
-        this.props.switchView(!this.props.placingOrder);
+    private _handleChange = (doNothing: boolean) => {
+        if (!doNothing) {
+            this.props.switchView(!this.props.placingOrder);
+        }
```
srkirkland: @jSylvestre how about instead something like `this.props.switchView('viewname')` instead of the do nothing.  Then it would be like how tabs or radios work.   So you'd change with this._switchView('browse') and then it would call the props.switchView which would take in the view name (or bool if you want to track it, but that limits us to binary)

PR #432, opened 2018-01-31
show spinner when submitting

### LINKED ISSUE #425: Add spinning wheel to Save & Review, and disable while it is running.

srkirkland:
closes #425


srkirkland:
@jSylvestre ah good point, I didn't think about the difference between 'can submit' and 'submitting'.  I'll change

PR #442, opened 2018-02-01
Lh/date sampled

### LINKED ISSUE #416: Add in "date sampled" date field for all orders.  required.
should go near commodity and just be a date field

laholstege:
closes #416

### DIFF
--- a/Anlab.Mvc/Client/components/OrderForm.tsx
+++ b/Anlab.Mvc/Client/components/OrderForm.tsx
@@ -311,10 +311,13 @@ export default class OrderForm extends React.Component<
                 commodity={commodity}
                 handleChange={this._handleChange}
               />
-              <DateSampled
-                  date={dateSampled}
-                  handleChange={this._handleChange}
-              />
+              <div className="form_wrap">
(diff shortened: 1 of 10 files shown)


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -20,6 +22,8 @@ import Summary from "./Summary";
 import { ITestItem, TestList } from "./TestList";
 import { ViewMode } from "./ViewMode";
 
+import "react-datepicker/dist/react-datepicker.css";
```
srkirkland: @laholstege This doesn't do anything right?  If so might as well dump it

Anlab.Mvc/Client/css/components.scss
```diff
@@ -1,4 +1,5 @@
 //images
+@import "~react-datepicker/dist/react-datepicker.css";
```
srkirkland: @laholstege just to see, try this without the ~ and instead use ./

want to see if that works

laholstege: @srkirkland doesn't work without the ~

Issue #427, opened 2018-01-30
Print page changes


srkirkland:
_Note, these will probably need to be discussed_

Basic idea -- add client info at the top of the page, including name, phone, email, etc.  Then add a section for "lab use only" that (ideally) just shows up in the print version. 

Also add lines for client signature.

Put the online order number in there somewhere

Then put in address where to submit samples to.

Attached is a sample screen that dirk sent

![image](https://user-images.githubusercontent.com/202753/35583301-5cc61dde-05a6-11e8-8504-e5213433f588.png)

PR #444, opened 2018-02-01
update print page with more info

srkirkland:
include lab only section, signature lines, and mail to info


srkirkland:
image

srkirkland:
closes #427

PR #455, opened 2018-02-06
Lh/client in i client info

### LINKED ISSUE #439: change color of new client button?
should not be blue unless the user has entered new client info

### LINKED ISSUE #441: clientID vs new client
if there is a clientID, disable the new client info button

change NewClientInfo class to just be client info, add string clientID field

add in copyEmail property (just for existing clients)

@jSylvestre also grab phone number along with client query

laholstege:
closes #441 closes #439 

still need #456


jSylvestre:
@laholstege One thing I noticed. If there is a default client id, it doesn't trigger the validation unless you change the field now.

PR #465, opened 2018-02-08
Lh/default emails

### LINKED ISSUE #417: PI's email address should default as one of 'send to' emails
Would require grabbing when we grab the client info.  Should not be editable.

laholstege:
closes #417


Anlab.Mvc/Client/components/ClientId.tsx
```diff
@@ -8,12 +8,14 @@ export interface IClientInfo {
...
     email: string;
+    copyEmail?: string;
+    subEmail?: string;
     phoneNumber: string;
 }
 
 interface IClientIdProps {
     //change one property of parent ClientInfo state
-    handleClientInfoChange: (key: string, value: string) => void;
+    handleClientInfoChange: (key: string[], value: string[]) => void;
```
srkirkland: @laholstege make a quick change to the comment since we are updating >=1 prop, and pluralize the prop names

PR #471, opened 2018-02-13
Lh/mail queue fix

laholstege:
fix css so pagination always works
show body in child row so it has less awkward spacing

image


jSylvestre:
@laholstege Did you want to add the failure reason to the title when it fails as well?
                        <i class="fa fa-times-circle" aria-hidden="true" title="@message.FailureReason"></i>
                        @:Failed

PR #475, opened 2018-02-14
Lh/account manager orders

laholstege:


Anlab.Mvc/Controllers/ReviewerController.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Anlab.Core.Data;
+using Microsoft.EntityFrameworkCore;
+using Anlab.Core.Models;
+using Microsoft.AspNetCore.Authorization;
+using AnlabMvc.Models.Roles;
+
+namespace AnlabMvc.Controllers
+{
+    [Authorize(Roles = RoleCodes.Admin)]
```
jSylvestre: @laholstege Need to add reports here too

PR #478, opened 2018-02-20
Jcs/save client name

### LINKED ISSUE #476: save client name in database
when an order is saved, when we save the clientID we also want to save the client name in a new field (even save the name if it's a new client).

When an order is received/updated in addRequestNumber, we also want to replace the clientID and client name field so they stay consistent.

Then, we update the lab orders table to show client name instead of user.

jSylvestre:
Closes #476

### DIFF
--- a/Anlab.Mvc/Controllers/LabController.cs
+++ b/Anlab.Mvc/Controllers/LabController.cs
@@ -57,7 +57,8 @@ namespace AnlabMvc.Controllers
                     RequestNum = c.RequestNum,
                     Status = c.Status,
                     ShareIdentifier = c.ShareIdentifier,
-                    Paid = c.Paid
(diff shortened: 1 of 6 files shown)


jSylvestre:
image

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -147,6 +148,7 @@ public IActionResult ListUsersOrders(string id)
                 return RedirectToAction("Orders");
             }
             order.ClientId = result.ClientId;
+            order.ClientName = "ERROR Client Name not found."; //Updated below if we find it
```
srkirkland: @jSylvestre change this to something like `"[Not Found]"` to make it simpler

PR #480, opened 2018-02-20
lab orders table changes

### LINKED ISSUE #477: lab admin order list changes
* ID --> Order ID
*ClientID --> Client ID

Add "hide complete" checkbox to the filters, which will be default checked and should hide all completed orders

laholstege:
closes #477


srkirkland:
@jSylvestre good point, I wasn't going to bother but i think you are right, we probably want the filter server side and then if the box is checked we can post back w/ includeCompleted=true or something

Issue #479, opened 2018-02-20
adding roles on Admin/Index doesnt seem to work


laholstege:
for example, if you make it so that the "Reviewers" tab only shows up if you are a reviewer, and then go to Admin/Index and add yourself, the layout doesn't change.

srkirkland:
yea, on sign-in the cookie you are issued has your roles and claims info.  So we'll need to re-issue a cookie when roles change, either by manually using the signinManager to handle cookies or by logging them out.  Either way, we still could always run into the problem where some other user was made an admin and will have to log out for it to take effect.  

I'd say when a role changes, if it's for the current user we can do `signinManager.SignIn(user, false)` in order to re-issue the cookie, and see if that works

PR #481, opened 2018-02-20
Lh/change RoleCodes.User to RoleCodes.LabUser and fix Admin/Clients

laholstege:
#479 

still have to decide best way to filter out admin/lab/reviewer from Admin/Clients


jSylvestre:
When this is deployed, just need to call
/Admin/CreateRole?role=LabUser

PR #496, opened 2018-02-28
Ability to Bcc Anlab on every email generated

### LINKED ISSUE #491: Bcc Labworks
All emails should bcc anlab

jSylvestre:
Closes #491


Anlab.Core/Models/EmailSettings.cs
```diff
@@ -10,5 +10,6 @@ public class EmailSettings
         public string Password { get; set; }
         public string Host { get; set; }
         public int Port { get; set; }
+        public string BccAnlab { get; set; }
```
srkirkland: @jSylvestre please change to Anlab or AnlabAddress or something. doesn't need to be bcc specific

PR #504, opened 2018-03-01
Have an indicator to show if the test item has an additional info pro…

jSylvestre:
…mpt.


srkirkland:
@jSylvestre screenshot? I'm hesitant to add another column for something that rarely happens

jSylvestre:
@srkirkland 
If we don't have this, it could be hard to track down without looking in the db:

image

srkirkland:
@jSylvestre ah ok, i thought it was on the client page.  for the admin side i don't care, go for it

PR #505, opened 2018-03-05
Jcs/lab use only image

jSylvestre:


jSylvestre:
Ugly on the phone:

image

srkirkland:
@jSylvestre we should probably just hide it unless you are printing anyway.  See if you can add that to the css and do this all at once, it should just be a one liner css class

jSylvestre:
@srkirkland Was just looking at the no print css:

```
        .noprint {
            display: none !important;
        }
```

PR #508, opened 2018-03-05
Jcs/shorten project

jSylvestre:
image


jSylvestre:
I wanted to use the string extensions class I created for the emails, so it is including that code too...

PR #514, opened 2018-03-06
Bypass Email to clients

jSylvestre:
image

### DIFF
--- a/Anlab.Mvc/Views/Lab/Confirmation.cshtml
+++ b/Anlab.Mvc/Views/Lab/Confirmation.cshtml
@@ -6,7 +6,7 @@
 }
 <div class="col">
     @Html.Partial("_OrderDetails")
-    <form asp-controller="Lab" asp-action="Confirmation" asp-route-id="@Model.Order.Id" method="post">
+    <form id="form" asp-controller="Lab" asp-action="Confirmation" asp-route-id="@Model.Order.Id" method="post">
         @Html.Hidden("confirm", true)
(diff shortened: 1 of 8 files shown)


jSylvestre:
image

Anlab.Mvc/Views/Lab/Confirmation.cshtml
```diff
@@ -6,7 +6,7 @@
 }
 <div class="col">
     @Html.Partial("_OrderDetails")
-    <form asp-controller="Lab" asp-action="Confirmation" asp-route-id="@Model.Order.Id" method="post">
+    <form id="form" asp-controller="Lab" asp-action="Confirmation" asp-route-id="@Model.Order.Id" method="post">
```
srkirkland: @jSylvestre this switch method is ok but it's also easy in the future if you just make 2 forms, then no javascript needed

jSylvestre: @srkirkland There is form data there. If there were 2 forms would that data be included in both?

srkirkland: @jSylvestre yea

srkirkland: @jSylvestre maybe not the best option for this specific situation

jSylvestre:
image

srkirkland:
@jSylvestre looks good 🚢

Issue #520, opened 2018-03-06
sample disposition


srkirkland:
every test needs sample disposition, which is basically a question asking how they want samples disposed.  It'll be like when we ask about how they like samples reported, except i won't be put into additional info, it'll need to be actually kept track of and shown on the email.

It'll be located near/next to date sampled.

laholstege:
@srkirkland are these from a list of options, or just a text box with user input?

PR #528, opened 2018-03-07
Lh/sample disposition

laholstege:
#520 
basic start for sample disposition. 
I wasn't sure if we wanted this to be preset options like the sample type questions are / what the options would be. 

also still need to add to emails, since I wasn't sure where we want it


Anlab.Mvc/Client/components/OrderForm.tsx
```diff
@@ -119,10 +123,10 @@ export default class OrderForm extends React.Component<
       placingOrder: true,
       project: "",
       quantity: null,
+      sampleDisposition: SampleDispositionOptions.dipose,
```
srkirkland: @laholstege type should be "dispose"

Anlab.Mvc/Client/components/SampleDisposition.tsx
```diff
@@ -0,0 +1,71 @@
...
+import Input from "./ui/input/input";
+
+interface ISampleDispositionProps {
+    sampleDispositionRef: (element: HTMLInputElement) => void;
+    handleChange: (key: string, value: any) => void;
+    disposition: string;
+}
+
+export const SampleDispositionOptions = {
+    dipose: "Dispose of my samples 30 days from report date.",
```
srkirkland: @laholstege ah, i see.  yea just type to dispose

PR #529, opened 2018-03-08
Warn if client id changes when WR number added

### LINKED ISSUE #524: Error or Warning?
If there is a clientId on the order, but when the order request number is added it is changed.

jSylvestre:
Closes #524


srkirkland:
@jSylvestre it'd kind be nice if it was a warning and not an error, but if you don't want to go through the trouble i don't blame you

PR #545, opened 2018-03-27
Copy Order

jSylvestre:
Working correctly.
Might get an error if someone not logged in tries to copy from the results page.


jSylvestre:
image

image

image

image

PR #547, opened 2018-04-05
Nicer page to maintain admin roles

jSylvestre:
Old:
image

New:
image

image

### DIFF
--- a/Anlab.Mvc/AnlabMvc.csproj
+++ b/Anlab.Mvc/AnlabMvc.csproj
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.BrowserLink" Version="2.0.2" />
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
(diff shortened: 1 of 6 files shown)


Anlab.Mvc/AnlabMvc.csproj
```diff
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.BrowserLink" Version="2.0.2" />
```
srkirkland: @jSylvestre why?

Anlab.Mvc/Views/Admin/EditAdmin.cshtml
```diff
@@ -0,0 +1,66 @@
...
+            <button class="btn btn-red" type="submit"><i class="fa fa-trash"></i> Admin</button>
+        </form>
+    }
+    else
+    {
+        <form method="post" asp-action="AddUserToRole" asp-route-userId="@Model.User.Id" asp-route-role=@RoleCodes.Admin asp-route-add="true" asp-antiforgery="true" style="margin: 3px">
+            <button class="btn btn-default" type="submit"><i class="fa fa-plus"></i> Admin</button>
+        </form>
+    }
+    @if (Model.IsLabUser)
```
srkirkland: @jSylvestre these are so close together, the only difference is one class and the route is false instead of true.  I'm thinking a quick partial would clean this up a ton

Anlab.Mvc/AnlabMvc.csproj
jSylvestre: @srkirkland Don't know.

jSylvestre:
@srkirkland Ok, take a look now. It is cleaner, but I'm not sure if this is exactly what you meant.

PR #549, opened 2018-04-09
Check the account expiry date and closed value

jSylvestre:


jSylvestre:
Tested with L-DOA113S and 3-CRU9033

jSylvestre:
Don't think it works with a sub account:
3-APSV016/FJHL2

PR #566, opened 2018-05-03
Make sure at send to sloth time it is upper case

jSylvestre:


jSylvestre:
https://github.com/ucdavis/Anlab/blob/master/Anlab.Mvc/Client/components/OrderForm.tsx#L557

PR #569, opened 2018-05-14
Jcs/admin search

jSylvestre:
image

image

### DIFF
--- a/Anlab.Mvc/Controllers/LabController.cs
+++ b/Anlab.Mvc/Controllers/LabController.cs
@@ -15,6 +15,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.EntityFrameworkCore;
 using Anlab.Core.Domain;
+using Anlab.Core.Extensions;
 using Anlab.Core.Services;
 using Serilog;
 using AnlabMvc.Helpers;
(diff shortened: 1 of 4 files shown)


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -420,6 +421,43 @@ public IActionResult Orders(bool showComplete)
...
+            return View();
+        }
+
+        [HttpPost]
+        public async Task<ActionResult> Search(string term)
+        {
+            Order order = null;
+            try
+            {
+                var shar = Guid.Parse(term);
```
srkirkland: @jSylvestre I'd much rather you use the TryParse methods instead of catching exceptions for non-exceptional behavior.  I think you can probably do this w/o any exceptions either by using tryParse methods or by using a Regex to figure out if it looks like a guid, integer, or other.

Anlab.Mvc/Views/Lab/Search.cshtml
```diff
@@ -0,0 +1,19 @@
+@{
+    ViewBag.Title = "Search";
+}
+
+
+<div class="col">
+    <form asp-controller="Lab" asp-action="Search">
+        <div class="form-horizontal">
```
srkirkland: @jSylvestre I think it kinda looks funny with the word wrap of the label, I think it would look better as a regular form instead of a horizontal one

jSylvestre:
image

PR #571, opened 2018-05-15
Jcs/tests20180514

jSylvestre:


Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -103,6 +103,12 @@ public IActionResult Orders(bool showComplete)
                 return NotFound();
             }
 
+            if (order.Status != OrderStatusCodes.Confirmed) //Have this here too because
```
srkirkland: @jSylvestre this is a strange comment

Anlab.Mvc/Models/Order/OverwriteOrderResult.cs
```diff
@@ -15,13 +15,14 @@ public OverwriteOrderResult()
...
+        public string ErrorMessage { get; set; }
         
         public string ClientId { get; set; }
         public int Quantity { get; set; }
         
         public decimal RushMultiplier { get; set; }
         public string LabworksSampleDisposition { get; set; }
 
-        public bool WasError => MissingCodes.Count > 0;
+        public bool WasError => MissingCodes.Count > 0 || ErrorMessage != null;
```
srkirkland: @jSylvestre maybe better to use isnullOrWhitespace in there, since empty string error message shouldn't count as an error

PR #585, opened 2018-05-24
remove regex on phones

laholstege:


Anlab.Core/Domain/User.cs
```diff
@@ -25,7 +25,6 @@ public class User : IdentityUser
         [Display(Name = "Client ID")]
         public string ClientId { get; set; }
 
-        [Phone]
```
jSylvestre: @laholstege I don't think we want to remove this one. This is the one related to a user when they are created. But I'll let @srkirkland have the final say.

srkirkland: @laholstege I don't really have a problem with removing this, I don't see the downside

PR #586, opened 2018-05-24
Return a null sop when not applicable

jSylvestre:


jSylvestre:
Ken's fixes to the query, plus deal with the null nicely

PR #603, opened 2018-07-10
Jcs/more new client info

jSylvestre:


jSylvestre:
@srkirkland React stuff. I think it is ok, but you should double check.

PR #604, opened 2018-07-10
Don't allow ucd emails to auth against google.

jSylvestre:
Note, ErrorMessage is not working.

### DIFF
--- a/Anlab.Mvc/AnlabMvc.csproj
+++ b/Anlab.Mvc/AnlabMvc.csproj
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.BrowserLink" Version="2.0.2" />
     <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0" />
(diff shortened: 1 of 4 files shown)


Anlab.Mvc/Controllers/AccountController.cs
```diff
@@ -290,6 +290,16 @@ public IActionResult ExternalLogin(string provider, string returnUrl = null)
                 return RedirectToAction(nameof(Login));
             }
 
+            if (info.LoginProvider.Equals("Google", StringComparison.OrdinalIgnoreCase))
+            {
+                var emailClaim = info.Principal.FindFirstValue(ClaimTypes.Email);
+                if (emailClaim.Contains("@ucdavis"))
+                {
+                    ErrorMessage = "Do not use Google to log in to a UC Davis account";
```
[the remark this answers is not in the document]
srkirkland: No I would think it would work since it's declared in the layout page.  Maybe it's cleared because of multiple redirects somehow (since error message only stays around in TempData for the next request)?

Anlab.Mvc/AnlabMvc.csproj
```diff
@@ -35,6 +35,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="2.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.0" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.BrowserLink" Version="2.0.2" />
```
jSylvestre: @srkirkland Should I delete this?

srkirkland: @jSylvestre probably doesn't hurt anything since it's a VS specific thing, but I would if you don't care about it

PR #605, opened 2018-07-11
Jcs/discount price

jSylvestre:
TODO:
1) Test that prices are correct -- Done
2) Test that prices are restored if they change the request number -- Done
3) Test that duplicate order restores prices -- Done

DB CHANGE!!!

### DIFF
--- a/Anlab.Mvc/Services/IOrderService.cs
+++ b/Anlab.Mvc/Services/IOrderService.cs
@@ -145,7 +145,18 @@ namespace AnlabMvc.Services
             order.ClientName = order.ClientName;
             order.JsonDetails = orderToCopy.JsonDetails;
             var orderDetails = order.GetOrderDetails();
-            var tests = CalculateTestDetails(order);
+            var tests = CalculateTestDetails(order).ToList();
+
+            var groupTestIds = tests.Where(a => a.Id.StartsWith("G-")).Select(a => a.Id).ToArray();
+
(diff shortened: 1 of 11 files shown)


jSylvestre:
@srkirkland  Duplicate copies tests that are included in the group. Should probably fix that, but maybe deploy this first?

Anlab.Mvc/Services/IOrderService.cs
```diff
@@ -144,7 +145,18 @@ private TestDetails[] CalculateTestDetails(Order order)
             order.ClientName = order.ClientName;
             order.JsonDetails = orderToCopy.JsonDetails;
             var orderDetails = order.GetOrderDetails();
-            var tests = CalculateTestDetails(order);
+            var tests = CalculateTestDetails(order).ToList();
+
+            var groupTestIds = tests.Where(a => a.Id.StartsWith("G-")).Select(a => a.Id).ToArray();
```
srkirkland: @jSylvestre since this is in-memory I think it'd be worth doing an ordinal string insensitive match, since I could totally see them messing up and doing a lowercase G.  You should just be able to add the comparer as an overload to the startsWith

srkirkland:
@jSylvestre also did you know in your comments/description you can make a checklist and then check it off later?

- [ ] you
- [ ] can
- [ ] do
- [ ] it

PR #606, opened 2018-07-12
Resend Failed Emails

jSylvestre:
Already manually deployed job

### DIFF
--- a/Anlab.Jobs.SendMail/Program.cs
+++ b/Anlab.Jobs.SendMail/Program.cs
@@ -55,7 +55,7 @@ namespace Anlab.Jobs.SendMail
 
 
             // Get all messages that we haven't tried to send yet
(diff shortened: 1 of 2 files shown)


Anlab.Jobs.SendMail/Program.cs
```diff
@@ -55,7 +55,7 @@ static void Main(string[] args)
 
 
             // Get all messages that we haven't tried to send yet
-            var messagesToSend = dbContext.MailMessages.Where(x => x.Sent == null).ToList();
+            var messagesToSend = dbContext.MailMessages.Where(x => x.Sent == null || x.Sent == false).ToList();
```
srkirkland: @jSylvestre go ahead and update the comment too since now we are getting unsent and failed messages

Issue #608, opened 2018-07-12
Pipe character on the list of prices public page


jSylvestre:
Replace the | character for the category with something that reads better.

PR #611, opened 2018-07-13
Jcs/setup fee for groups

jSylvestre:


jSylvestre:
Closes #608

Anlab.Core/Models/TestItemPrices.cs
```diff
@@ -35,7 +35,7 @@ public decimal SetupPrice
         {
             get
             {
-                if (Nonrep)
+                if (Nonrep && !Id.StartsWith("G-", StringComparison.OrdinalIgnoreCase))
```
srkirkland: @jSylvestre seems ok, though we are doing this check in a lot of places.  Next time you come around to it maybe throw together a helper to figure out if the ID represents a group so we don't have to keep doing the startsWith

PR #615, opened 2018-07-16
Add missing tests

jSylvestre:
If an order's copy of all the tests is missing one that it pulled down from anlab, try to add that to the order's all tests with the current price.
This will only work if the test exists in our DB though.

### DIFF
--- a/Anlab.Mvc/Models/Order/OverwriteOrderResult.cs
+++ b/Anlab.Mvc/Models/Order/OverwriteOrderResult.cs
@@ -12,11 +12,13 @@ namespace AnlabMvc.Models.Order
         {
             SelectedTests = new List<TestDetails>();
             MissingCodes = new List<string>();
+            MissingTestsToAdd = new List<TestItemModel>();
         }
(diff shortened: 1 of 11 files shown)


Anlab.Mvc/Models/Order/OverwriteOrderResult.cs
```diff
@@ -12,11 +12,13 @@ public OverwriteOrderResult()
...
             SelectedTests = new List<TestDetails>();
             MissingCodes = new List<string>();
+            MissingTestsToAdd = new List<TestItemModel>();
         }
         public IList<TestDetails> SelectedTests { get; set; }
 
         public IList<TestItemModel> BackedupTests { get; set; }
-        public IList<String> MissingCodes { get; set; }
+        public IList<TestItemModel> MissingTestsToAdd { get; set; }
+         public IList<String> MissingCodes { get; set; }
```
srkirkland: @jSylvestre i guess i missed this earlier but in the future use lowercase `string` so it uses the primitive correctly

PR #622, opened 2018-07-19
Jcs/edits20180719

jSylvestre:


Anlab.Mvc/Views/Analysis/Index.cshtml
```diff
@@ -186,6 +186,6 @@
                     </ul>
                 </div>
             </div>
-        </div>
+        </div>*@
```
srkirkland: @jSylvestre just delete we don't need this anymore

PR #629, opened 2018-07-20
Logout

jSylvestre:
@srkirkland This works (would need to pass in the app settings for the CAS URL). But I feel that the logout stuff should be setup in the startup file. But, I'm not sure how to do that.
When I use @jpknoll 's code, I get exceptions like

`InvalidOperationException: No IAuthenticationSignOutHandler is configured to handle sign out for the scheme: Google`

Note: It looks like CAS has the logout redirect disabled so that wouldn't work anyway.


srkirkland:
@jSylvestre yea i'm ok with this as a workaround but I would bet strongly that when we setup the auth providers in startup.cs we can specify logout config there and then hopefully you won't get that invalid operation anymore.

PR #630, opened 2018-07-24
Ping that checks both DBs

jSylvestre:


jSylvestre:
@srkirkland If you don't like that injected into the home controller, I can create a new controller just for the ping.

PR #637, opened 2018-08-02
Cal's script

jSylvestre:


Anlab.Mvc/Views/Shared/Layout.cshtml
```diff
@@ -160,5 +160,13 @@
     <script src="~/dist/root.js" asp-append-version="true"></script>
 
     @RenderSection("Scripts", required: false)
+    <script async src=“https://www.googletagmanager.com/gtag/js?id =UA-652647-50“></script>
```
jSylvestre: @srkirkland This was the only addition. Changes above were just auto formatting

srkirkland: @jSylvestre According to google it should be the first thing under <head>.  It might work here, who knows.

srkirkland: @jSylvestre also it looks like you have a space in the URL that i don't think should be there

PR #638, opened 2018-08-03
Jcs/move email templates src

jSylvestre:


jSylvestre:
@srkirkland Not sure if you like how I added these to the solution, but it makes it easier to make changes and copy the dest changes to our template in the MVC views. Those changes are in the second commit

PR #642, opened 2018-08-07
Force GRNDONLY and DM55 codes not to have a setup price.

### LINKED ISSUE #640: GRNDONLY  and DM55 should not have a setup fee
Grind isn’t showing a setup fee, but I didn’t know about GRNDONLY and DM55

jSylvestre:
Move to string extension
closes #640

### DIFF
--- a/Anlab.Core/Extensions/StringExtensions.cs
+++ b/Anlab.Core/Extensions/StringExtensions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Globalization;
 
 namespace Anlab.Core.Extensions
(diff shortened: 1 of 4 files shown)


jSylvestre:
Waiting to hear back from Dirk

Anlab.Core/Extensions/StringExtensions.cs
```diff
@@ -13,5 +14,29 @@ public static string SafeToUpper(this string value)
 
             return value.ToUpper(CultureInfo.InvariantCulture);
         }
+
+        public static bool ClearOutSetupPrice(this string value)
+        {
```
srkirkland: @jSylvestre you can do this in one line `return string.Equals("GRINDONLY", value, StringComparison.OrdinalIgnoreCase)`

Anlab.Core/Extensions/StringExtensions.cs
```diff
@@ -13,5 +14,29 @@ public static string SafeToUpper(this string value)
...
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public static bool IsGroupTest(this string value)
+        {
+            if (value.StartsWith("G-", StringComparison.OrdinalIgnoreCase))
```
srkirkland: @jSylvestre I'd use string.StartsWith() in order to avoid dealing w/ nulls

Anlab.Core/Extensions/StringExtensions.cs
jSylvestre: @srkirkland Good point, originally Dirk had told me two values, but now that there is only one that is better.

Anlab.Core/Extensions/StringExtensions.cs
jSylvestre: @srkirkland What do you mean? Just remove the if?

PR #664, opened 2018-08-17
Special Truncate

jSylvestre:
I want to truncate only if the total length is greater than the max, but I want to truncate in the middle (The project name, not the request #)

### DIFF
--- a/Anlab.Core/Services/SlothService.cs
+++ b/Anlab.Core/Services/SlothService.cs
@@ -9,7 +9,6 @@ using System;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
-using System.Runtime.InteropServices.ComTypes;
 using System.Threading.Tasks;
 using Anlab.Core.Extensions;
 using Serilog;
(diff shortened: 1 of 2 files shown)


Anlab.Core/Services/SlothService.cs
```diff
@@ -50,12 +51,14 @@ public async Task<SlothResponseModel> MoveMoney(Order order)
                 Log.Error("Sloth Token missing");
             }
 
+            var xxx = $"{order.Project.SpecialTruncation((order.RequestNum.Length + 3), 40)} - {order.RequestNum}";
```
srkirkland: Dump this I don't see it used

Anlab.Core/Services/SlothService.cs
```diff
@@ -9,6 +9,7 @@
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Runtime.InteropServices.ComTypes;
```
srkirkland: Why would you need this?

PR #693, opened 2018-10-11
Catch specific email exception

jSylvestre:


jSylvestre:
image

Anlab.Jobs.SendMail/Program.cs
```diff
@@ -67,24 +68,42 @@ static void Main(string[] args)
...
 
                     message.Sent = true;
                     message.SentAt = DateTime.UtcNow;
                     counter++;
                 }
+                catch (SmtpFailedRecipientException failedRecipientEx)
+                {
+                    Log.Error(failedRecipientEx.Message);
+                    message.FailureCount = 5; //Say it failed 5 times so it doesn't try to send again. Otherwise need to change the db.
+                    var messageWithOrder = dbContext.MailMessages.Include(i => i.Order).SingleOrDefault(a => a.Id == message.Id);
```
srkirkland: @jSylvestre might not be worth it but seems like this could be done more efficiently by just projecting out the orderId, like 

dbContext.MailMessages.Where(a=>a.Id == message.Id).Select(a => a.Order.Id).SingleOrDefault();

PR #705, opened 2018-10-17
Jcs/confirmation page20181016

### LINKED ISSUE #699: formatting on order confirmation page
Try to make some of the client info and basic info in a justified new column layout

jSylvestre:
Closes #699

### DIFF
--- a/Anlab.Mvc/Extensions/StringExtensions.cs
+++ b/Anlab.Mvc/Extensions/StringExtensions.cs
@@ -64,9 +64,13 @@ namespace AnlabMvc.Extensions
             var phoneNumberUtil = PhoneNumbers.PhoneNumberUtil.GetInstance();
             var phoneNumber = phoneNumberUtil.ParseAndKeepRawInput(value, "US");
             var phone = phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.NATIONAL);
-            if (phone == "+1 NA")
+            if (!phoneNumberUtil.IsValidNumberForRegion(phoneNumber, "US"))
             {
                 return phoneNumber.RawInput;
+            }
+            if (phone == "+1 NA") //Just in case catch
+            {
+               return phoneNumber.RawInput;
             }
 
             return phone;
--- a/Anlab.Mvc/Extensions/StringExtensions.cs
+++ b/Anlab.Mvc/Extensions/StringExtensions.cs
@@ -73,10 +73,6 @@ namespace AnlabMvc.Extensions
 
                 return phoneNumber.RawInput;
             }
-            if (phone == "+1 NA") //Just in case catch
-            {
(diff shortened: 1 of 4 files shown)


Anlab.Mvc/Extensions/StringExtensions.cs
```diff
@@ -53,5 +54,22 @@ public static bool IsEmailValid(this string value)
...
+        public static string FormatPhone(this string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                return value;
+            }
+            var phoneNumberUtil = PhoneNumbers.PhoneNumberUtil.GetInstance();
+            var phoneNumber = phoneNumberUtil.ParseAndKeepRawInput(value, "US");
+            var phone = phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.NATIONAL);
+            if (phone == "+1 NA")
```
srkirkland: @jSylvestre seems to me it'd be better to check if the phone number is valid instead of checking if the formatted string == "+1 NA".  I'm sure the parsed result has a valid or similar prop on it

Anlab.Mvc/Extensions/StringExtensions.cs
```diff
@@ -53,5 +54,22 @@ public static bool IsEmailValid(this string value)
...
         }
+
+        public static string FormatPhone(this string value)
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                return value;
+            }
+            var phoneNumberUtil = PhoneNumbers.PhoneNumberUtil.GetInstance();
+            var phoneNumber = phoneNumberUtil.ParseAndKeepRawInput(value, "US");
```
srkirkland: why are we assuming US?  They get lots of international numbers

Anlab.Mvc/Extensions/StringExtensions.cs
```diff
@@ -53,5 +54,22 @@ public static bool IsEmailValid(this string value)
...
+            var phone = phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.NATIONAL);
```
srkirkland: Let's make sure this works with international numbers, like a german number might be 001-541-754-3010

Anlab.Mvc/Extensions/StringExtensions.cs
jSylvestre: @srkirkland If I use null, it fails.
And above where I format it, it looks nicer if I use National (US)

Anlab.Mvc/Extensions/StringExtensions.cs
jSylvestre: image

jSylvestre: With is valid for region:

PR #725, opened 2019-02-06
Hot Fix Disable PhoneNumber formatter

jSylvestre:
Wrap phone number format in a try catch. But restarting the app seems to have fixed the error.

### DIFF
--- a/Anlab.Mvc/Extensions/StringExtensions.cs
+++ b/Anlab.Mvc/Extensions/StringExtensions.cs
@@ -57,6 +57,7 @@ namespace AnlabMvc.Extensions
 
         public static string FormatPhone(this string value)
         {
+            return value;
             if (string.IsNullOrWhiteSpace(value))
             {
                 return value;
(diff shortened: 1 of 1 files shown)


Anlab.Mvc/Extensions/StringExtensions.cs
```diff
@@ -57,24 +57,33 @@ public static bool IsEmailValid(this string value)
...
-                return value;
-            }
-            var phoneNumberUtil = PhoneNumbers.PhoneNumberUtil.GetInstance();
-            var phoneNumber = phoneNumberUtil.ParseAndKeepRawInput(value, "US");
-            var phone = phoneNumberUtil.Format(phoneNumber, PhoneNumberFormat.NATIONAL);
-            if (!phoneNumberUtil.IsValidNumberForRegion(phoneNumber, "US"))
+            try
             {
-                if (phoneNumberUtil.IsValidNumber(phoneNumber))
+                if (string.IsNullOrWhiteSpace(value))
```
srkirkland: @jSylvestre no point in putting this in the try/catch.  But i guess not a problem either

PR #730, opened 2019-05-03
Lh/uc payment

laholstege:
image
hides the selection on editing
we still only keep isUcdAccount, not any info on which UC they choose

### DIFF
--- a/Anlab.Mvc/Client/components/PaymentUcSelection.tsx
+++ b/Anlab.Mvc/Client/components/PaymentUcSelection.tsx
@@ -1,0 +1,103 @@
+import "isomorphic-fetch";
+import * as React from "react";
+import { getOptions } from "../util/ucAccountsHelper";
+import Input from "./ui/input/input";
+
+export interface IPayment {
+  clientType: string;
+  account?: string;
+  accountName?: string;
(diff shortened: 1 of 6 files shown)


jSylvestre:
@laholstege I'll fix the spec, and test it on Monday

Anlab.Mvc/Client/components/PaymentUcSelection.tsx
```diff
@@ -0,0 +1,109 @@
...
+                        maxLength={50}
+                        onChange={this.props.handleAccountChange}
+                        onBlur={this.props.lookupAccount}
+                        inputRef={this.props.ucAccountRef}
+                    />
+                {this.props.payment.accountName}
+                </div>
+                {this.state.ucName === "UCB" && this._renderDetails(options["UCB"])}
+                {this.state.ucName === "UCSF" && this._renderDetails(options["UCSF"])}
+                {this.state.ucName === "UCD" && this._renderDetails(options["UCD"])}
```
srkirkland: What's the point of all of these in-render IFs instead of just a `this._renderDetails()` which checks this.state.ucName and returns the proper input?  You might have to add in a check if options(name) is null but that seems way cleaner than writing them out

PR #748, opened 2019-05-07
Some extra info on the error pages

### LINKED ISSUE #737: Update generic error message
Add some language like "Please try again.  IF the problem persists, contact anlab@ucdavis.edu" or similar.  Possibly send them to the FAQ or suggest for them to try again in incognito?

jSylvestre:
Close #737


jSylvestre:
> Looks good but do we have that email in a config somewhere?

Yes, but we also have it in many other places including markdown. I just use the appSetting for the email jobs

PR #754, opened 2019-05-14
Hot Fix

jSylvestre:


jSylvestre:
This was failing because I was trying to show billing info like Billing Contact, but it was a ICD order being deleted with admin override. (We don't need to notify account about those I think.)

Issue #741, opened 2019-05-07
30 day disposal warning


srkirkland:
•	Specific EMAILS from TOPS
o	Generate a 30-day disposal warning
	Clients choose [save] or [dispose] samples
	TOPS knows when the report was sent / finalized.
	When the administrator pushed the button, a list of all clients (with their work request numbers) is generated whose work is past the 30-day grace period.  Administrator would then select all or some of the list to send as a warning.

jSylvestre:
Looking at the data, there are 4 possible values for the LabworksSampleDispositionField:
Blank - Maybe an error
D - Dispose?
R - Return?
S - Save

Have the query list all those that are not D or R ?

jSylvestre:
Look at Peaks bulk edit for multi select

srkirkland:
@jSylvestre this ready to close?

PR #758, opened 2019-05-16
Jcs/disposal email

jSylvestre:


jSylvestre:
Issue #741

Anlab.Mvc/Controllers/LabController.cs
```diff
@@ -826,6 +827,79 @@ public async Task<ActionResult> Search(string term)
...
+
+            await _dbContext.SaveChangesAsync();
+
+            Message = "Email Generated";
+
+            return RedirectToAction("MailQueue", "Admin", new {id = order.Id});
+        }
+
+        [HttpPost]
+        public async Task<IActionResult> GenerateDisposalEmails(string ids)
```
srkirkland: @jSylvestre these two are basically the same thing so either move the core logic into a private method or more likely just use GenerateDisposalEmails for everything even if it's just one they are generating

PR #763, opened 2019-05-24
Why did this change?

jSylvestre:


jSylvestre:
@jpknoll I loaded the PR from new new computer onto my old computer, updated node
did a npm rebuild
then this was changed.
Should it have?

srkirkland:
BTW this change looks legit -- I think one of our dependencies change minor versions and then your install did an auto-update and pulled in this extra stuff.

PR #778, opened 2019-09-19
cleanup package and yarn locks

srkirkland:
using npm 6.11.3, full cache purge and lockfile regen


srkirkland:
well, this didn't work.  which means our package lock can't be recreated, which isn't great.  probably some minor version thing?

PR #779, opened 2019-09-19
remove yarn lock and update package-lock

srkirkland:


srkirkland:
FYI, anlab uses yarn.  So we don't want to dump the yarn.lock file, that causes problems.  The csproj runs `npm install` though.  So we are in a strange place.  Going to close this an just leave alone for a while

PR #791, opened 2020-01-02
Update Index.cshtml

benningd54:
Under "Plants" heading was labeled wrong.


benningd54:
Line 87.
Changed "Storage and Containers"   to  "Method Summaries"

And that's all!

-db

PR #813, opened 2020-04-23
Set up CI with Azure Pipelines

srkirkland:


srkirkland:
should be ready to go now @jSylvestre take a look if you want when you get a chance.   pipeline that it makes is https://dev.azure.com/ucdavis/Anlab/_build?definitionId=19&_a=summary

PR #829, opened 2020-08-26
What the new URLs will be

jSylvestre:
Just to document what needs to change

### DIFF
--- a/Anlab.Mvc/appsettings.json
+++ b/Anlab.Mvc/appsettings.json
@@ -22,6 +22,8 @@
     "StorageContainerName": "anlabupload", //Dev
     "CyberSourceUrl": "https://testsecureacceptance.cybersource.com/pay",
     "FinancialLookupUrl": "https://kfs.ucdavis.edu/kfs-prd/remoting/rest/fau",
+    "FinancialLookupUrlNewProd": "https://financials.api.adminit.ucdavis.edu/fau",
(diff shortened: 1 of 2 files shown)


Anlab.Mvc/appsettings.json
```diff
@@ -22,6 +22,8 @@
     "StorageContainerName": "anlabupload", //Dev
     "CyberSourceUrl": "https://testsecureacceptance.cybersource.com/pay",
     "FinancialLookupUrl": "https://kfs.ucdavis.edu/kfs-prd/remoting/rest/fau",
+    "FinancialLookupUrlNewProd": "https://financials.api.adminit.ucdavis.edu/fau",
+    "FinancialLookupUrlNewUat": "https://financials-uat.api-np.adminit.ucdavis.edu/fau",
```
sprucely, line 26: Do we want to keep putting urls into source control, or should we start marking these as [external] and place the actual urls in 1password?

jSylvestre: Generally, but not always, we use External for secrets. Nothing secret about this.
I'm not going to merge this, just use it as a reminder for when we add an app setting,

sprucely: > I'm not going to merge this, just use it as a reminder for when we add an app setting,

Oh, now I get it.

PR #830, opened 2020-08-26
Use new account to send anlab's emails

jSylvestre:
Need to also change app setting for email

### DIFF
--- a/Anlab.Core/Services/MailService.cs
+++ b/Anlab.Core/Services/MailService.cs
@@ -38,7 +38,8 @@ namespace Anlab.Core.Services
         {
             Log.Information($"Email Host: {_emailSettings.Host}");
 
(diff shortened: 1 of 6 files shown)


Anlab.Core/Services/MailService.cs
```diff
@@ -106,5 +109,10 @@ public void SendFailureNotification(int orderId, string failureReason)
                 Log.Error($"Error sending email for Failure reason. OrderId {orderId} Exception {ex.Message}" );
             }
         }
+
+        public bool DisableSend()
```
srkirkland, line 113: Change the name to like `IsSendDisabled` because I thought this was a method to actually disable send.  In fact maybe you could just do it as a prop like so:
`public bool IsSendDisabled { get { return ... } }`

PR #836, opened 2020-09-14
Jcs/try to update emails

jSylvestre:


jSylvestre:
image

jSylvestre:
When I actually sent it, it looked like this in my email (missing extra space between <p>

image

jSylvestre:
@srkirkland When I generated the email. it ended stripping out all the formatting.
We should look at replacing the emails for anlab...

srkirkland (approved):
changes seem ok.  And yes I agree it might be about time to change over the emails to something newer

PR #843, opened 2020-09-30
Update Index.cshtml

benningd54:
Update to give information for Smoke Taint


jSylvestre (changes requested):
We don't use direct links like that. 
I'll fix it up in a couple minutes.

jSylvestre (approved):

benningd54:
Am I supposed to do something now?
   The only button I see is to "Review changes"  and there are three
options there.

-db

jSylvestre:
> Am I supposed to do something now? The only button I see is to "Review changes" and there are three options there.

@benningd54 Just merge it when you are ready.

PR #844, opened 2020-10-01
Update Index.cshtml

benningd54:


jSylvestre (approved):
FYI, you may want to move some of the text out of the link, or surround more of it.
This is what it looks like when you hover over it:

image

jSylvestre (changes requested):
You have extra on text

jSylvestre (approved):

PR #859, opened 2021-01-04
Update Index.cshtml

benningd54:
Remove Special Announcement Box


jSylvestre (approved):
Didn't test it, but looks good to me.

PR #870, opened 2021-02-24
Add elastic logging

sprucely:
Verified that web and both jobs are logging to elastic.


srkirkland (approved):
looks great but I'm a little scared about the package-lock.json changing.  As long as we check that clicking around the order page still works after this change i'm ok leaving it, but i'm always scared updating JS lib versions.

sprucely:
There aren't actually any lib version updates this time in `packages-lock.json`; just some new transitive dependencies.

A bit of pointing and clicking didn't reveal any problems on the orders page, but I'll wait for @jSylvestre to also do some testing.

jSylvestre (approved):
I didn't see any issues, and didn't test jobs.

Don't forget to update the app config in azure before deploying

PR #888, opened 2021-06-30
Changes to get sass to work with latest node

jSylvestre:


sprucely (approved):
Worked for me, running node v12.18.3

srkirkland (approved):
Looks good but let's keep an eye on the build/deploy. It's possible we might need to update our pipelines script.

PR #890, opened 2021-08-03
Update Index.cshtml

benningd54:
Update announcement to advertise new methods


jSylvestre (approved):
Didn't test, but looks ok to me.

PR #898, opened 2021-09-15
Update Index.cshtml

benningd54:
Note about long turnaround times


jSylvestre (approved):
For some reason that looked like you were missing a div, but looks fine when I run it:

image

PR #913, opened 2022-01-26
Update rush-request.md

benningd54:
Update Rush charges


jSylvestre (approved):
You may or may not want to add a period to the last sentence. It was missing it before.

image

PR #922, opened 2022-03-18
Update Index.cshtml

benningd54:
Upcoming Holiday CLosure Info


jSylvestre (approved):
When it is merged and deployed, you can look at the staging slot to make sure it looks right.
Then slack me and I can swap staging with prod.

PR #929, opened 2022-04-21
CRA upgrade

srkirkland:
Upgrade to CRA, which is really just the order page.   Created a few orders and everything appears to be working.  We do get console warnings in dev, the most concerning of which are about deprecated/removed react lifecycle events used by the bootstrap and datepicker libs, though these components still seem to be working fine so perhaps it doesn't affect us.

Tests are completely broken (karma/enzyme doesn't support react17) so submitting as a draft while I fix or consider different options.


Anlab.Mvc/Views/Order/Edit.cshtml
```diff
@@ -7,20 +7,18 @@
...
-<div id="react-app" class="col"></div>
+<div id="root" class="col"></div>
 
-@section Scripts
-{
+@section ScriptVariables {
     <script>
         App = {};
         App.orderData = @Json.Serialize(Model);
-        App.orderId = @Model.Order.Id;
```
jSylvestre, line 17: I think this is causing a new order to be created on edit

jSylvestre (changes requested):

jSylvestre (approved):

Anlab.Mvc/ClientApp/src/components/ClientId.tsx
```diff
@@ -117,12 +117,21 @@ export class ClientId extends React.Component<IClientIdProps, IClientIdInputStat
         this._validateModal();
     }
 
+    private _notEmpty = (value: string) => {
```
jSylvestre, line 120: Nice

sprucely (approved):
Looked at it last week, so just went over the most recent commits. Looks good.

PR #931, opened 2022-04-25
Jcs/version and warnings

jSylvestre:


sprucely (approved):

srkirkland (changes requested):
Let's try again but without changing all the file formatting

PR #944, opened 2022-07-12
WIP alerts

jSylvestre:


jSylvestre:
image

Anlab.Mvc/Controllers/AlertsController.cs
```diff
@@ -0,0 +1,130 @@
...
+        }
+
+        // GET: AdminAnalysis/Create
+        public ActionResult Create()
+        {
+            return View();
+        }
+
+        [HttpPost]
+        public async Task<ActionResult> Create(SystemAlert systemAlert)
```
srkirkland, line 36: Instead of making the Domain model have fake non-required fields and other stuff like empty created/updated, I think it'd just be simpler to create a viewModel here for this create method.  Just something that can get `IsActive`, markdown, description, etc the required form fields.  you don't have to use the domain SystemAlert for both DB and model binding

Anlab.Mvc/Controllers/AlertsController.cs
```diff
@@ -0,0 +1,130 @@
...
+            
+            return View(alert);
+        }
+        
+        [HttpPost]
+        public async Task<ActionResult> ToggleActive(int id)
+        {
+            var alert = await _dbContext.SystemAlerts.SingleAsync(x => x.Id == id);
+            alert.IsActive = !alert.IsActive;
+            _dbContext.Update(alert);
```
srkirkland, line 75: not needed, save changes knows what changes need to be saved

Anlab.Mvc/Controllers/AlertsController.cs
```diff
@@ -0,0 +1,130 @@
...
+                var alertToUpdate = await _dbContext.SystemAlerts.SingleAsync(x => x.Id == systemAlert.Id);
+                alertToUpdate.IsActive    = false; //Force to false when editing.
+                alertToUpdate.Markdown    = systemAlert.Markdown;
+                alertToUpdate.Danger      = systemAlert.Danger;
+                alertToUpdate.Updated     = DateTime.UtcNow;
+                alertToUpdate.Description = systemAlert.Description;
+
+                var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().UseBootstrap().Build();
+                alertToUpdate.Content = Markdown.ToHtml(alertToUpdate.Markdown, pipeline);
+                _dbContext.Update(alertToUpdate);
```
srkirkland, line 105: not needed

Anlab.Mvc/Views/Alerts/Edit.cshtml
```diff
@@ -0,0 +1,57 @@
...
+                </div>
+            </div>
+
+            <div class="form-group">
+                <label asp-for="Markdown" class="col-md-2 control-label"></label>
+                <div class="col-md-10">
+                    <textarea asp-for="Markdown" class="form-control" rows="10"> </textarea>
+                    <span asp-validation-for="Markdown" class="text-danger"></span>
+                    <div>
+                        <small>To see basic commands on how markdown works, google Markdown cheat sheet. Not all options are supported with the markdown library we use but most are. You may also add in html, but we recommend that you don't.</small>
```
srkirkland, line 27: will html work?  i would think the model binder would auto reject html as a potential security issue

jSylvestre: Yes, it just goes in as raw. Same thing the other place we use this.
I only tested with <br/> but it is expected behavior with markdown from what I can see.

jSylvestre:
> Does `<textarea/>` default to a mono-spaced font? I think it does. Otherwise that would be helpful for markdown. I don't have anything else to add on top of Scott's comments.

Not sure, but this is what our other page that does markdown is using.

PR #958, opened 2023-02-06
Aggie Account

jSylvestre:
User's account in Profile. DB Change!


Anlab.Core/Services/AggieEnterpriseService.cs
```diff
@@ -0,0 +1,208 @@
...
+                rtValue.IsValid = data.GlValidateChartstring.ValidationResponse.Valid;
+
+                if (!rtValue.IsValid)
+                {
+                    foreach (var err in data.GlValidateChartstring.ValidationResponse.ErrorMessages)
+                    {
+                        rtValue.Messages.Add(err);
+                    }
+                }
+                rtValue.Details.Add(new KeyValuePair<string, string>("Entity", $"{data.GlValidateChartstring.SegmentNames.EntityName} ({data.GlValidateChartstring.Segments.Entity})"));
```
jSylvestre, line 70: I probably don't need these details, but going to leave in there for now in case we want to display any of this info on the lab details page.

srkirkland (approved):
🚢 looks good to me.  Only slight concern is when linking the finjector script it's the same code copy/paste into several files -- so if we ever update to v2 you have to do it in several places.  I don't have a good solution for it but something to consider for later.

PR #995, opened 2023-05-09
View Deleted Orders

jSylvestre:
Only shows them if no active order with search info was found.


jSylvestre:
image

srkirkland (approved):
looks ok but we have to actually make sure they can't be acted upon.  I'm guessing since the other methods will respect the query params that's not a huge issue but potentially it's worth hiding all the action buttons.

PR #1001, opened 2023-05-17
Check Sloth URL

jSylvestre:


srkirkland (commented):
Be better to have it once in a helper method but the idea is fine

PR #1003, opened 2023-05-22
Jcs/email report

jSylvestre:


Anlab.Mvc/Controllers/ReviewerController.cs
```diff
@@ -88,7 +88,48 @@ public async Task<IActionResult> Details(int id)
...
+            {
+                query = query.Where(a => a.CreatedAt <= model.End.Value.Date.AddDays(1).FromPacificTime());
+            }
+            if(model.OrderAction == "Confirmed")
+            {
+                query = query.Where(a => a.Subject == "Work Order Confirmation" || a.Subject.StartsWith("Work Request Confirmation"));
+            }
+            else
+            {
+                query = query.Where(a => a.Subject.StartsWith("Work Request Finalized"));
```
srkirkland, line 125: this could get pretty gnarly if we don't put any indexes on subject.   especially since email tables get large.  Let's keep an eye on it and make sure it doesn't take a long time running on prod.

srkirkland (approved):

PR #1004, opened 2023-05-24
Docusign support

srkirkland:
Adds in new docusign flow based on impersonation of a TBD service account.  When order is confirmed, user is redirected to `PendingSignature` which creates a new docusign envelope generated via a new `Document` HTML view, which we can modify to look like anything.  They are then redirected to docusign to complete the signature.  Once successful, it will redirect back to `SignatureCallback` and the envelope info will be attached and the order will be marked completed as before.

Also added a link to the view order page which shows a link to view the signed document if it exists.

DRAFT because we'll want to change up the wording and style of both the current confirmed order page as well as the signing document.


Anlab.Mvc/Services/DocumentSigningService.cs
```diff
@@ -0,0 +1,229 @@
...
+        {
+            throw new ArgumentException("Order not found");
+        }
+
+        // first, get auth token
+        var authToken = AuthenticateWithJwt();
+
+        var user = GetAccountInfo(authToken);
+
+        var account = user.Accounts[0]; // TODO: what if there are multiple accounts?  need to select one somehow perhaps
```
jSylvestre, line 58: @srkirkland 
Get all user accounts and look for is_default?
https://developers.docusign.com/platform/auth/reference/user-info/

srkirkland: Yes that seems like a good thing to do and the right solution.

jSylvestre (approved):
I have a pr #1009 for the todo

Anlab.Mvc/Services/DocumentSigningService.cs
```diff
@@ -0,0 +1,254 @@
...
+            AnchorXOffset = "20",
+        };
+
+        var dateSigned = new DateSigned
+        {
+            AnchorString = "**date_1**",
+            AnchorUnits = "pixels",
+            // AnchorYOffset = "10",
+            AnchorXOffset = "20",
+        };
```
sprucely, line 241: Should the signature location somehow be made configurable/dynamic? Or is this good enough for the few documents it's used on?

sprucely (approved):
Looks good to me. Just a couple comments. And had to lookup some oauth stuff. Was there any reason to choose impersonation over delegation?

PR #1008, opened 2023-06-07
Create View

jSylvestre:
Prod and Test DBs should have this view

### DIFF
--- a/Anlab.Core/Data/ApplicationDbContext.cs
+++ b/Anlab.Core/Data/ApplicationDbContext.cs
@@ -96,11 +96,11 @@ namespace Anlab.Core.Data
                 .OnDelete(DeleteBehavior.Cascade);
 
             builder.Entity<ReviewerOrderView>()
-                .ToTable("ReviewerOrders");
+                .ToView("ReviewerOrders");
 
             //Probably don't need this here because the name is the same as the table name
             builder.Entity<HistoricalSalesView>()
-                .ToTable("HistoricalSalesView");
+                .ToView("HistoricalSalesView");
 
 
             base.OnModelCreating(builder);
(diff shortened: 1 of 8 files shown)


Anlab.Core/Data/ApplicationDbContext.cs
```diff
@@ -96,6 +98,10 @@ protected override void OnModelCreating(ModelBuilder builder)
             builder.Entity<ReviewerOrderView>()
                 .ToTable("ReviewerOrders");
 
+            //Probably don't need this here because the name is the same as the table name
+            builder.Entity<HistoricalSalesView>()
+                .ToTable("HistoricalSalesView");
```
srkirkland, line 103: You can (should??) use `.ToView()` here.  I'm not sure what it does in practice, but I assume it keeps an internal mapping that is read/query only.

Anlab.Mvc/Controllers/ReviewerController.cs
```diff
@@ -129,6 +130,92 @@ public async Task<IActionResult> EmailList(DateTime? start, DateTime? end, strin
...
+            if (model.Start != null)
+            {
+                query = query.Where(a => a.DateFinalized >= model.Start.Value.Date.FromPacificTime());
+            }
+            if (model.End != null)
+            {
+                query = query.Where(a => a.DateFinalized <= model.End.Value.Date.AddDays(1).FromPacificTime());
+            }
+
+            var results = await query.ToListAsync(); //Grab all the results for the date range for processing
```
srkirkland, line 158: If you use .ToView() it probably won't matter, but doing `AsNoTracking()` will save you some memory overhead.

PR #1009, opened 2023-06-08
Get the first account, default preferred

jSylvestre:


Anlab.Mvc/Services/DocumentSigningService.cs
```diff
@@ -56,7 +56,7 @@ public async Task<string> SendForEmbeddedSigning(int orderId)
 
         var user = GetAccountInfo(authToken);
 
-        var account = user.Accounts[0]; // TODO: what if there are multiple accounts?  need to select one somehow perhaps
+        var account = user.Accounts.OrderByDescending(a => a.IsDefault).First(); // TODO: what if there are multiple accounts?  need to select one somehow perhaps
```
srkirkland, line 59: Just change to `user.Accounts.Single(a => a.IsDefault)`

srkirkland (changes requested):

Anlab.Mvc/Services/DocumentSigningService.cs
jSylvestre: Ok. I just wasn't sure every user had one that was default.
Then debugging, I saw it was your account :)

jSylvestre: @srkirkland Ugg, it is a string.

srkirkland: @jSylvestre oh yea, but don't use string.equals -- EF Core doesn't know how to use that so it doesn't translate like you think.  Just use `Single(a => a.IsDefault == "true");`.  SQL is case insensitive anyway.

srkirkland: for more details if curious: https://learn.microsoft.com/en-us/ef/core/miscellaneous/collations-and-case-sensitivity#translation-of-built-in-net-string-operations

srkirkland (approved):

PR #1019, opened 2023-06-22
Option to allow Certificate to be shown

jSylvestre:
Change the parameter to showCert?


sprucely (approved):
Didn't test, but code looks fine.

PR #1042, opened 2023-11-08
Update Index.cshtml

benningd54:
Update to include Phenolic preparation


jSylvestre:
List of Method summaries differences would just be from my local data
image

jSylvestre (approved):
You might want to do the uppercase however in a string tag instead

PR #1077, opened 2025-02-06
Get if order is finished in labworks

### LINKED ISSUE #1075: Investigate
Is it possible to check if the work order is finished and has had an email sent?

Finished — that is, within LabWorks the “()” acode has a value (initials) in it.
Email sent — view TOPS record to see if there is an email record
Then, if there is one without the other, there should be a flag set and an alert.

jSylvestre:
Close #1075


jSylvestre:
image

image

jSylvestre:
image

image

image

PR #1092, opened 2025-04-23
Populate printed name

jSylvestre:
image


jSylvestre:
No, Don't do this.
It is used for manually receiving and needs to be blank.

PR #1094, opened 2025-04-30
change email addr to anlab@notify.ucdavis.edu

srkirkland:
now we should be able to send via sparkpost if we change the login info to use sparkpost smpt settings too.


jSylvestre (approved):
Had to hide whitespace...

srkirkland:
yuck i just did it on github directly and it killed the whitespace.  but oh well

PR #1124, opened 2025-08-07
font change on btn and textarea height

cydoval:


jSylvestre:
image

jSylvestre:
This is what it looked like in the old BS3 system:
image

